PR #2, opened 2016-03-28
Database mapping macro

spalladino:
Mapping macro, similar to YAML or JSON mapping, to populate an instance of a class from a resultset.


bcardiff:
This depends on Hash been ordered, which may change in the near future https://github.com/crystal-lang/crystal/issues/2331

spalladino:
I don't see where it depends on the hash being sorted, as it matches the column name with the mapping key (exactly like the YAML mapping does)

spalladino:
As for the travis check failure, it seems to be a change in Crystal 0.14 (it was working fine in 13)

bcardiff:
I meant that https://github.com/bcardiff/crystal-db/pull/2/files#diff-89ed3becdd76dae7d10377988fac60a5R91 iterates the fields in certain order.

bcardiff:
@spalladino could you rebase & update for the removal of `read?`. Maybe the last changes in JSON mapping to support nillable/unions offers some guidance to deal better with unions.

src/db/mapping.cr
```diff
@@ -0,0 +1,136 @@
...
+                  %rs.read(Union({{value[:type]}} | Nil))
+                {% else %}
+                  %rs.read({{value[:type]}})
+                {% end %}
+          {% end %}
+          else
+            {% if strict %}
+              raise DB::MappingException.new("unknown result set attribute: #{col_name}")
+            {% else %}
+              %rs.read(Nil)
```
bcardiff: %rs.read  (without arguments) maybe ?

src/db/mapping.cr
```diff
@@ -0,0 +1,136 @@
...
+      def {{key.id}}=(_{{key.id}} : {{value[:type]}} {{ (value[:nilable] ? "?" : "").id }})
+        @{{key.id}} = _{{key.id}}
+      end
+
+      def {{key.id}}
+        @{{key.id}}
+      end
+    {% end %}
+
+    def self.from_rs(%rs : DB::ResultSet)
```
bcardiff: this should go away, we can merge it as is, and remove it when adapting to the new query_one/query_all api.
agreed?

spalladino: I'd like to keep this to be consistent with the JSON and YAML mapping APIs

PR #9, opened 2016-06-28
Require ResultSet to just implement `read`, optionally implementing `read(T.class)`. Fixes #5

### LINKED ISSUE #5: Data type extensibility
Allow each driver to extend the data types allowed.

asterite:
The `scalar` method now also returns a union, and one can cast it to the needed type.


asterite:
Implementations:
* sqlite3: https://github.com/crystal-lang/crystal-sqlite3/pull/9
* mysql: https://github.com/crystal-lang/crystal-mysql/pull/1

src/db/result_set.cr
```diff
@@ -55,30 +54,14 @@ module DB
...
-    # The result is one of `DB::TYPES`.
-    abstract def column_type(index : Int32)
+    # Reads the next column value
+    abstract def read
 
-    def read(t)
-      read?(t).not_nil!
+    # Reads the next column value as a **type**
+    def read(type : Class)
+      type.cast(read)
```
bcardiff: * shouldn't this base implementation simplify the code of `dummy_driver#read(T)` ?
* does Class match structs and unions?

asterite: Yes, a lot. I think the dummy driver can just have all the values as an array of DB::Any. I'll try :-)

asterite: Yes, Class matches any type, even union types.

asterite: @bcardiff I don't know how to easily simplify the specs, I can see `DummyResultSet` receives the query and splits it by commas. I'd leave it like this for now so `read(Class)` is also tested.

asterite:
I've just updated this PR with some additions. Sample code:

```cr
DB.open "sqlite3://%3Amemory%3A" do |db|
  db.exec "create table contacts (name string, age integer)"
  db.exec "insert into contacts values (?, ?)", "John Doe", 30

  # `read(*types)` to read multiple values at once
  db.query "select name, age from contacts" do |rs|
    rs.each do
      name, age = rs.read(String, Int32)
      pp name, age
      # name # => "John Doe"
      # age  # => 30
    end
  end

  # `query_one` to read a single row
  # This raises if there are no rows, or more than one row
  name, age = db.query_one "select name, age from contacts", &.read(String, Int32)
  pp name, age
  # name # => "John Doe"
  # age  # => 30

  # similar to `query_one`, but returns nil if there are no rows (still raises if there are many rows)
  result = db.query_one? "select name, age from contacts", &.read(String, Int32)
  pp result
  pp typeof(result)
  # result # => {"John Doe", 30}
  # typeof(result) # => (Tuple(String, Int32) | Nil)

  db.exec "insert into contacts values (?, ?)", "Foo Bar", 40

  # `query_all` to read an array of values
  people = db.query_all "select name, age from contacts order by age desc", &.read(String, Int32)
  pp people, typeof(people)
  # people         # => [{"Foo Bar", 40}, {"John Doe", 30}]
  # typeof(people) # => Array(Tuple(String, Int32))
end
```

This fixes #10 in a much better way.

And the nice thing about `query_one`, `query_one?` and `query_all` is that the returned value depends on the block, so one can create an instance of a class as a result and, for example, get an array of proper classes instead of an array of tuples. Or use named tuples, etc.

I think with this we could easily migrate `crystal-pg` to this and have the same functionalities, even with a nicer syntax.

/cc @bcardiff @waj @spalladino

bcardiff:
1. `#scalar` should be available in query methods. is shorter and useful. but it can be implemented in QueryMethods and not in Statement for sure.

2. Can't we have `query_one sql, T1, T2 ... TN`  and be type safe? It would be a better syntax IMO. Or is there something I am missing? I can't imagine other than a read what I will want to have. Maybe the things we would miss there is the named tuple support. 

But other than that I think we can merge this. :-) I like it.

asterite:
@bcardiff The problem is that `query_one` also accepts the arguments, like:

```cr
db.query_one "select name from contacts where id = ? limit 1", 123, &.read(String)
#                                                              ^-- argument 
```

so it would be hard to distinguish between arguments and types.

bcardiff:
if only something like this would work...

```
db.query_one {String, String}, "select name, email from contacts where id = ? limit 1", 123
db.query_one "select name, email from contacts where id = ? limit 1", 123, as: {String, String}
```

But well, let's keep the `&.read(String)` until a better alternative comes.

asterite:
@bcardiff The first form, `query {String, String}, "select ...", args` is how you do things in `crystal-pg` [right now](https://github.com/will/crystal-pg#typed-querying). But I'm not sure it reads very well.

I like the `as` form, however, it reads much better: first the query, then the bind arguments, then the expected type.

I've updated the PR with those forms, and updated the sample code in the comment above.

asterite:
My only "problem" or question with the `as` argument is this:

```cr
# This is OK, returns a string and an int32
db.query_one "...", as: {String, Int32}

# What about this?
db.query_one "...", as: Tuple(String, Int32)
```

Right now the second way doesn't compile, because that ends up calling `rs.read(Tuple(String, Int32))` and you can't read a tuple from a column. Maybe that's fine. But one should remember to always use a tuple as an `as` argument, not a tuple type (probably not a problem because it's easier to write `{...}` than `Tuple(...)`)

Of course if a DB driver allows you to read a tuple from a single column, the second form will work.

asterite:
I also left the block versions in all cases, because you can then do:

```cr
point = db.query_one("select lat, lng from locations where id = ?", 123) { |rs| Point.new(*rs.read(Float64, Float64)) }
point # => Point(...)
```

bcardiff:
It's nice how `{String, Int32}` is of type `Tuple(String:Class, Int32:Class)` and then we can instantiate the method for that specific type. :-)

PR #12, opened 2016-09-01
Feature/pool

### LINKED ISSUE #7: Connection pool
Should be parametrisable via connection string URI. Ensure it is still possible to quickly setup a new connection without creating a full pool, for quick open-exec-close operations. Must keep track of "busy" and "free" connections.

bcardiff:
Fixes #7 

Adds a connection pool to manage connection. All connection pool options are configurable by URI query string. Example: `DB.open("mysql://user:pass@host/dbname?initial_pool_size=2&max_pool_size=10")`

Options:
* `initial_pool_size` initial number of connections to be opened (defaults 1)
* `max_pool_size` maximum number of connections to be opened (defaults 1)
* `max_idle_pool_size` keep no more than this amount of idle/not used connections in the pool (defaults 1)
* `checkout_timeout` fail to connect to the db if no connection turns to be available after this amount of seconds. (defaults 5.0 seconds)
* `retry_attempts` if an established connection is lost allow the pool to reconnect, retry only this number of times before failing (defaults 1)
* `retry_delay` amount of seconds to wait before each retry attempt to reconnect (defaults 1.0 seconds)

Drivers implementors can override how parameters are fetch. Or force them for example if only single connection is supported. See `Driver#connection_pool_options`.

A `PoolStatement` is introduced that represents statements that the user does not care in which actual connection will be executed. Hence, it can try to reconnect based on reconnect policy of the pool.

`Pool#retry` receives a block that will be re-executed if the block raises a `DB::ConnectionLost`.
This is used in `PoolStatement#exec` and `#query` to ensure an actual statement over a working connection is returned.

There are some outstanding issues regarding releasing connection reference once they are closed that will require callbacks or weak references.

Implementation assumes single thread since there is no mutex added in the pool.

It might be worth adding an overload to connect to the DB like `DB#connect(config_uri)` that will yield a connection (not pooled) in order to allow quick single connection access to a database. Probably in another PR.


src/db/connection.cr
```diff
@@ -18,31 +14,29 @@ module DB
...
 
     abstract def build_statement(query) : Statement
 
     protected def do_close
-      @statements_cache.each do |_, stmt|
-        stmt.close
-      end
+      @statements_cache.each_value &.close
       @statements_cache.clear
     end
```
spalladino: Remove connection from pool when explicitly closing

src/db/pool.cr
```diff
@@ -0,0 +1,195 @@
...
+    private def can_increase_idle_pool
+      @available.size < @max_idle_pool_size
+    end
+
+    private def pick_available
+      @available.first
+    end
+
+    private def wait_for_available
+      timeout = TimeoutHelper.new(@checkout_timeout.to_f64, ->{ @availability_channel.send nil })
```
spalladino: Use a different channel for the timeout helper, and then `select` on it and the availability channel

sdogruyol:
@bcardiff can't wait to have this merged in 👍

Xosmond:
@bcardiff Do you have an example of how to use it please?

RX14:
@Xosmond you just use the library as usual and it works in the background

bcardiff:
@Xosmond also check https://crystal-lang.org/docs/database/connection_pool.html it offers some code snippets and description of behavior.

Xosmond:
@bcardiff Thanks mate

PR #23, opened 2016-10-22
default max_pool_size to 0 = unlimited

bcardiff:
We've been discussing what should be a default max_pool_size for database drivers.

Before this the default was 1. Which will be a common bottle neck for concurrent applications.

Other languages/libs defaults are:
- .net=100
- java jdbc=100
- ruby activerecord=2
- go=∞ (0 == ∞)

This PR changes the default to 0 meaning ∞ as in go.

cc: @asterite @waj

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## (unreleased)
 
+* Changed default connection pool size limit is now 0 (unlimited).
+
 * Fixed allow new connections right away if pool can be increased
 
 ## v0.2.0 (2016-10-20)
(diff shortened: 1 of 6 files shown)


RX14:
I think 100 being default is probably about right. Having too many db connections to most databases is a really bad idea, and having a sane default limit instead of asking the user to configure it is a good idea. 

Having 0 be unlimited is fine, but I believe having unlimited the default is a hidden error waiting to happen.

bcardiff:
@RX14 Any fixed limit seems artificial for concurrent applications. I agree it could be a hidden error waiting, but if the bottleneck turns to be an artificial limit I think that will be harder to discover.

PR #24, opened 2016-11-22
support Time and Bool in base driver

drujensen:
Hi,  I added `Time` and `Bool` to the DB types.

### DIFF
--- a/spec/dummy_driver.cr
+++ b/spec/dummy_driver.cr
@@ -163,6 +163,29 @@ class DummyDriver < DB::Driver
       read(String).to_f64
     end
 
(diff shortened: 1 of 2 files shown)


spec/dummy_driver.cr
```diff
@@ -163,6 +163,29 @@ class DummyDriver < DB::Driver
       read(String).to_f64
     end
 
+    def read(t : Time.class)
```
bcardiff: These changes to the dummy_driver are not been used. For the sake of just-compiling the code could be just smaller.

If this code is used for some specs that is a different story for sure.

PR #25, opened 2016-12-04
Support non prepared statements

### LINKED ISSUE #3: Support non prepared statements

bcardiff:
fixes #3 

Breaking changes:
* `Connection#build` is renamed to `Connection#build_prepared_statement`.
* `Connection#build_unprepared_statement` is added.

Added:
* add a `prepared_statements=true|false` (default: `true`) to the connection uri. This is forwarded to `Database` and `Connection` as `#prepared_statements? : Bool`
* `Database` and `Connection` has also `#prepared` and `#unprepared` methods that return `QueryMethod` instances to query/exec prepared and unprepared statements despite the connection uri.
* `Connection#prepared(sql : String)` and `Database#prepared(sql: String)` should be used to create prepared statements explicitly that will be executed later. Note: there is a cache by sql statement so same query is not prepared twice on connection. This is not new, but proper interface was pending.

A `SessionMethods` module is added to share code between `Database` and `Connection` regarding how to build statements of the different kinds. When transactions are implemented this will be helpful also. This module is generic due to implementation details: `Database` works with `PoolStatement` and `Connection` works with `Statement`.

cc: @spalladino @asterite


src/db.cr
```diff
@@ -120,17 +120,29 @@ module DB
   private def self.build_database(uri : URI)
     Database.new(driver_class(uri.scheme).new, uri)
   end
+
+  # :nodoc:
+  def self.fetch_bool(params : HTTP::Params, name, default : Bool)
+    if value = params[name]?
+      value.underscore == "true"
```
spalladino: Underscore?

bcardiff: oops. changed to downcase and also I validate the value now.

src/db/pool_unprepared_statement.cr
```diff
@@ -0,0 +1,21 @@
...
+      super
+    end
+
+    protected def do_close
+      # unprepared statements do not need to be release in each connection
+    end
+
+    # builds a statement over a real connection
+    private def build_statement
+      @db.pool.checkout.unprepared.build(@query)
```
spalladino: I'd rather call `build_unprepared_statement` directly here, to avoid creating the extra intermediate struct

spalladino:
LGTM. We seriously need to discuss documentation, though :-P

src/db/pool_unprepared_statement.cr
bcardiff: since it is a struct I would bet llvm will inline that. We could even add the @[AlwaysInline] maybe. I prefer use public api if possible. build_(un)prepared_statements are to be defined by drivers but I wouldn't suggest to use them as public api: they are too verbose.

spalladino: Good point on the inlining!

shayneoneill:
Out of curiosity, whats the use case for this? The problem with an uprepared statement is that people start using them, aka PHP in the 90s and early 2000s and start generating SQL injection hell. Is there a  solid reason to support using them, or is there something I'm missing here?

bcardiff:
@shayneoneill prepared statement consume resources in the database. With long living connections this could be a problem. There is also no profit in using prepared statements if the statements changes  often like in dynamic queries. The author of crystal-pg also vouched for using non prepared statements as per his experience in heroku.

SQL injection is independent of prepared/unprepared statements. You can always concatenate ugly strings and submitted. The difference is if you want to keep a handle to that statements for later re-execution without the need to build an execution plan again.

PR #26, opened 2016-12-04
added types for int8, int16, Time for mysql driver

crisward:
I get an error on mysql without these.
Related to this - https://github.com/crystal-lang/crystal-mysql/pull/15

### DIFF
--- a/src/db.cr
+++ b/src/db.cr
@@ -69,7 +69,7 @@ module DB
   # Types supported to interface with database driver.
   # These can be used in any `ResultSet#read` or any `Database#query` related
   # method to be used as query parameters
-  TYPES = [Nil, String, Int32, Int64, Float32, Float64, Bytes]
+  TYPES = [Nil, String, Int8, Int16, Int32, Int64, Float32, Float64, Bytes, Time]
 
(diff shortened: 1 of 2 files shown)


bcardiff:
Time will arrive to DB::Any. I am a bit hesitant regarding Int8/Int16.

Int8/Int16 and specially UInt8/16 are mainly for interfacing with C.
Some dbs  support even unsigned, but I am not they all should be added here.

@crisward I am not following why without this PR you get errors in mysql. Can you share some dump/log?

crisward:
Error here - https://gist.github.com/crisward/5b228395b8b7e14d7669d59ed9b32faf

bcardiff:
@crisward Int8/16 stories should be fixed at crystal-mysql@master. check crystal-lang/crystal-mysql#18 . 

I though this would been covered by [this spec](https://github.com/crystal-lang/crystal-mysql/blob/a1db0df999cd6247b946745893ed98979cc113a9/spec/driver_spec.cr#L99-L114) but the case without unions in the exec args didn't raise the issue you point in the error.

If that is ok, then we could close this issue since mysql can extend the type to be used outside DB::Any.

crisward:
Thanks, that fixes it. I'll close.

PR #27, opened 2016-12-13
Feature/nested transactions

bcardiff:
Add support for transactions using BEGIN/ROLLBACK/END statements by default.
Add support to mimic nested transactions using SAVEPOINTS.
Transactions statements are executed using non prepared statements (overridable if needed per driver).

Some checks ensure there are no overlapping transactions per connection or per parent transaction.

```crystal
db.transaction do |tx|
  tx.connection.query ....
  # raise exception to rollback exception will bubble   
  # or tx.rollback or tx.commit
  # or raise DB::Rollback exception will not bubble

  tx.transaction do |inner|
    inner.connection.query ...
    # same as tx for rollback
  end

  other_inner = tx.begin_transaction
  # use other_inner transaction without block
  other_inner.commit
end
```


bcardiff:
@will after merging https://github.com/will/crystal-pg/pull/66 this will work also in crystal-pg and I will PR the specs of https://github.com/bcardiff/crystal-pg/tree/feature/nested-transactions like in mysql and sqlite.

Your input is welcome for sure.

spec/transaction_spec.cr
```diff
@@ -0,0 +1,158 @@
...
+    cnn = uninitialized DB::Connection
+    with_dummy do |db|
+      db.transaction do |tx|
+        cnn = tx.connection
+        cnn.scalar "1"
+        db.pool.is_available?(cnn).should be_false
+      end
+      db.pool.is_available?(cnn).should be_true
+    end
+  end
```
spalladino: I'd add another spec here: ensure that a connection is not returned to the pool if a nested transaction is commited/rolledback, but there is still an outer active one.

src/db/begin_transaction.cr
```diff
@@ -0,0 +1,26 @@
...
+        # TODO check if `unless tx.closed?` should be added.
+        # If it is not added, when an exception occurs after the
+        # transaction is closed (explicit rollback/commit)
+        # the call to rollback will trigger an exception.
+        # Since the code will be in between `#transaction(&block)`
+        # seems reasonable to raise an error since it might trigger
+        # a design flaw. Transaction might been commited already.
+        # Maybe we should wrap e in another exception that clarifies
+        # this scenario.
+        tx.rollback
```
spalladino: +1 to adding `unless tx.closed?`

If we're at this point, there has already been an exception thrown from the user's code, which will bubble and be catched by the client code. If the tx is already closed, it means that he manually commited or rolled it back, so it makes no sense to me to try and do that again, triggering an additional exception. The design flaw is exposed by the original exception, there's no need for another one.

@asterite maybe you want to break the tie? :-P

src/db/error.cr
```diff
@@ -17,4 +17,7 @@ module DB
     def initialize(@connection)
     end
   end
+
+  class Rollback < Exception
+  end
```
spalladino: I'm a bit confused about Error vs Exception here. Could you explain?

src/db/connection.cr
```diff
@@ -42,10 +44,60 @@ module DB
...
+
+    protected def create_transaction : Transaction
+      TopLevelTransaction.new(self)
+    end
+
     protected def do_close
       @statements_cache.each_value &.close
       @statements_cache.clear
       @database.pool.delete self
     end
```
spalladino: Is it safe to close a connection without rolling back its transactions? I assume it is, but just checking.

src/db/connection.cr
```diff
@@ -21,10 +21,12 @@ module DB
   abstract class Connection
     include Disposable
     include SessionMethods(Connection, Statement)
+    include BeginTransaction
 
     # :nodoc:
     getter database
     @statements_cache = StringKeyCache(Statement).new
+    @transaction = false
```
spalladino: Why not store the `TopLevelTransaction` object itself here, so we can keep a reference to it from the Connection?

src/db/transaction.cr
```diff
@@ -0,0 +1,118 @@
...
+      close!
+    end
+
+    def rollback
+      close!
+    end
+
+    private def close!
+      raise DB::Error.new("Transaction already closed") if closed?
+      close
```
spalladino: Maybe there is an `abstract def close` missing here? Or is that covered by Disposable?

src/db/transaction.cr
```diff
@@ -0,0 +1,118 @@
...
+      super
+    end
+
+    protected def do_close
+      connection.release_from_transaction
+    end
+
+    def begin_transaction : Transaction
+      raise DB::Error.new("There is an existing nested transaction in this transaction") if @nested_transaction
+      @nested_transaction = true
```
spalladino: Same as with `@transaction` at `Connection`

src/db/transaction.cr
```diff
@@ -0,0 +1,118 @@
...
+
+    # :nodoc:
+    def create_save_point_transaction(parent : Transaction) : SavePointTransaction
+      # TODO should we wrap this in a mutex?
+      previous_savepoint = @savepoint_name
+      savepoint_name = if previous_savepoint
+                         previous_savepoint.succ
+                       else
+                         # random prefix to avoid determinism
+                         "crystal_#{Random.rand(10_000)}_00001"
```
spalladino: I'd increase 10_000 a bit more. [Just in case](https://en.wikipedia.org/wiki/Birthday_problem).

src/db/transaction.cr
```diff
@@ -0,0 +1,118 @@
...
+    # :nodoc:
+    def release_from_nested_transaction
+      @nested_transaction = false
+    end
+  end
+
+  class SavePointTransaction < Transaction
+    getter connection : Connection
+
+    def initialize(@parent : Transaction,
```
spalladino: -1 to argument wrap. But I'm just being picky.

src/db/transaction.cr
```diff
@@ -0,0 +1,118 @@
...
+    def commit
+      close!
+    end
+
+    def rollback
+      close!
+    end
+
+    private def close!
+      raise DB::Error.new("Transaction already closed") if closed?
```
spalladino: I'm wondering if we maybe should let just the DB fail in this case, rather than try to play smarter and throw the exception ourselves. However, I can't think of any scenario where we think the tx is `closed?` but it actually isn't.

src/db/error.cr
bcardiff: DB::Error are used when the origin of the exception was an interaction with the database.
Rollback is initiated by the user.

src/db/connection.cr
bcardiff: It was simpler to just store the bool.
We can have a `TopLevelTransaction` or `WeakRef(TopLevelTransaction)` but if doing that we should deal with explicit management of what will happen when the connection or parent transaction is closed IMO.

src/db/connection.cr
bcardiff: Connection can always go down, so yes, they are not committed by deafault.

src/db/transaction.cr
bcardiff: I think is safe to track this state and offer better errors in this scenario.
Also it is not clear that the driver can query the status of the transaction, so I will keep this design.

src/db/transaction.cr
bcardiff: it is covered in `Disposable` which is defined also in DB ;-)

src/db/transaction.cr
bcardiff: We could go expose top_level_transaction in Transaction to avoid the wrap. That will allow us to remove `SavePointTransaction#create_save_point_transaction` but the SavePoint needs: 
* parent to release/unlock
* top_level to increase new savepoint name
* connection to exec

src/db/transaction.cr
bcardiff: `1_000_000` ? 

the usage of the prefix is also to when dealing with a pool, so not all savepoints of all connections are named equally.

When the same connection is returned to the pool _it should_ have no longer outstanding transactions. dsl and the design enforce that when using the public api. So the risk is about dirty state between checkouts in the same connection. For long living connections the numbers will be reused, but again, between different connections the names can be the same without problems to the db.

We can prefix with the `connection.object_id` also...

src/db/transaction.cr
spalladino: I wasn't thinking about _querying_ the status, but rather simply issuing the `close` statement to the DB, and catch the "already closed" exception. However, either option is ok for me.

src/db/transaction.cr
spalladino: The `object_id` approach sounds good to me

src/db/transaction.cr
bcardiff: Well, I prefer caution here. If a TopTransaction ref somehow leak at hold a reference to an open connection it could commit/rollback a not intended transaction. It should not happen, unless bugs, but let's keep this.

src/db/connection.cr
bcardiff: Well, tracking the object force either a refactor of how the BEGIN/CREATE SAVEPOINT is done OR a :nodoc: setter is needed. Because we need to flag the connection as in-transaction **before** the sql command is sent.

I think I would prefer to keep bools for now, [the required change](https://gist.github.com/bcardiff/bd4fbfed985439880d60a9f1332e4d2b) is not that nice. Either way it should not impact to drivers implementors since they don't need to inherit Transaction class, or if done, they should call base constructor.

src/db/transaction.cr
bcardiff: done.

src/db/transaction.cr
bcardiff: done.

spec/transaction_spec.cr
bcardiff: done.

spalladino:
LGTM!

PR #38, opened 2017-01-16
Add Database#checkout, Connection#release for non block connection use

bcardiff:
Refactor the checkout/release code for connection to allow non block connection checkout from DB.

### DIFF
--- a/spec/database_spec.cr
+++ b/spec/database_spec.cr
@@ -114,6 +114,41 @@ describe DB::Database do
     end
   end
 
+  it "should checkout different connections until they are released" do
+    DummyDriver::DummyConnection.clear_connections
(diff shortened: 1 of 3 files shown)


spec/database_spec.cr
```diff
@@ -114,6 +114,28 @@ describe DB::Database do
...
+
+      the_first_cnn.release
+      db.pool.is_available?(the_first_cnn).should be_true
+      db.pool.is_available?(the_second_cnn).should be_false
+
+      db.checkout.should eq(the_first_cnn)
+      the_first_cnn.release
+      the_second_cnn.release
+    end
+  end
```
spalladino: I'd add a spec that checks that if one of the checked-out connections executes a statement and closes the underlying resultset, then the connection is **not** returned to the pool.

bcardiff: done.

spalladino:
LGTM, feel free to merge!

PR #40, opened 2017-02-14
Allow nilable types in DB.mapping

RX14:
Previously, using `DB.mapping(foo: String?)` instead of `DB.mapping(foo: {type: String, nilable: true})` would fail with a nil assertion error.

JSON.mapping does the cast the same way as this.

### DIFF
--- a/src/db/mapping.cr
+++ b/src/db/mapping.cr
@@ -65,6 +65,10 @@ module DB
     {% end %}
 
     {% for key, value in properties %}
+      {% value[:nilable] = true if value[:type].is_a?(Generic) && value[:type].type_vars.map(&.resolve).includes?(Nil) %}
+    {% end %}
+
+    {% for key, value in properties %}
       @{{key.id}} : {{value[:type]}} {{ (value[:nilable] ? "?" : "").id }}
 
       def {{key.id}}=(_{{key.id}} : {{value[:type]}} {{ (value[:nilable] ? "?" : "").id }})
@@ -131,7 +135,7 @@ module DB
         {% elsif value[:default] != nil %}
           @{{key.id}} = %var{key.id}.is_a?(Nil) ? {{value[:default]}} : %var{key.id}
         {% else %}
-          @{{key.id}} = %var{key.id}.not_nil!
+          @{{key.id}} = %var{key.id}.as({{value[:type]}})
         {% end %}
       {% end %}
     end
--- a/spec/mapping_spec.cr
+++ b/spec/mapping_spec.cr
@@ -29,6 +29,13 @@ class MappingWithNilables
   })
 end
 
+class MappingWithNilTypes
+  DB.mapping({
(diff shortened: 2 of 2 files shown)


bcardiff:
I agree that `foo: String?` should be supported. I am surprised the change didn't require a `Union({{value[:type]}})` since that serves for jumping to the type grammar inside the statements grammar.

Either way, we would need some additions in `mapping_spec.cr` for this.

Would you add them? Checking that both `foo: String?` and `foo: String | Nil` works as you expect?

src/db/mapping.cr
```diff
@@ -65,6 +65,10 @@ module DB
     {% end %}
 
     {% for key, value in properties %}
+      {% value[:nilable] = true if value[:type].is_a?(Generic) && value[:type].type_vars.map(&.resolve).includes?(Nil) %}
```
RX14: The types come through as `Generic` ast nodes represented as `::Union(String, ::Nil)` so it requires quite a song and dance to ascertain if the types are nilable. Is there any way to improve this @asterite?

bcardiff: @RX14 check https://github.com/crystal-lang/crystal/blob/master/src/json/mapping.cr#L128

RX14: @bcardiff That doesn't work in macro code, only in crystal code.

RX14:
Anything left to do here?

sdogruyol:
Up 👍

bcardiff:
@RX14 I've just merge this and extended to support `Nil | T`, `T | Nil` syntax and not only `T?`.

Thanks!

sdogruyol:
Great job 🎉

RX14:
Why on earth would `Type?` and `Type | Nil` be treated differently when they're type aliases? All this code required just to work out if a type is a union and contains nil seems like a huge deficiency in the macro system.

bcardiff:
They are different at the grammar level

* `T?` is always a Type
* `T | Nil` is a call to `T.|(Nil)` 
* `Union(T | Nil)` is always a type (I think)

In the hash `{col1: T | Nil}`, the value of `col1` is a call ast node. There is no "nice" way to enforce the `col1` value should be a type.

In JSON.mapping the workaround to detect this is less painful because there is no need to update the `:nilable` per variable due to how the expanded code works.

I agree is not ideal, but supporting `T?` but neither `T | Nil` nor `Nil | T` seems worst.

RX14:
If you could do something like `{% parse_type(value[:type].to_s) %} ` or similar it would be very helpful for solving this problem. It would take a string and parse the contents as if it were type grammar, and return the resulting ASTNode.

bcardiff:
Maybe just `{% as_type(value[:type]) %}` is enough and be able to apply thing like `{% as_type(value[:type]).nilable? %}`.

Feel free to open a issue in crystal to discuss that feature.

PR #44, opened 2017-03-07
Implement DB.connect to create non pooled connections

### LINKED ISSUE #13: Add DB.connect to create connections without connection pool
`DB.connect` should return a `DB::Connection` that is not managed by a pool.
Refactor Database & Pool dependency of Connection.

bcardiff:
Fixes #13

I already have the changes to mysql, sqlite and pg drivers.
Merging this will bump a minor version change in crystal-db.

Consumers of database drivers should not be affected by this change.

### DIFF
--- a/src/db.cr
+++ b/src/db.cr
@@ -113,12 +113,42 @@ module DB
     end
   end
 
+  # Opens a connection using the specified *uri*.
+  # The scheme of the *uri* determines the driver to use.
+  # Returned connection must be closed by `Connection#close`.
+  # If a block is used the connection is yielded and closed automatically.
+  def self.connect(uri : URI | String)
+    build_connection(uri)
+  end
+
+  # ditto
+  def self.connect(uri : URI | String, &block)
+    cnn = build_connection(uri)
+    begin
+      yield cnn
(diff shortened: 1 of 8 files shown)


bcardiff:
cc: @will, @spalladino, @rdp in case any of you want to have word on this. This allows the user to create straight connections without a pool.

* `DB.open` remains to be pooled.
* `DB.connect` will create a single independent connection.

will:
Cool

src/db.cr
```diff
@@ -113,12 +113,42 @@ module DB
...
 
+  # Opens a connection using the specified *uri*.
+  # The scheme of the *uri* determines the driver to use.
+  # Returned connection must be closed by `Connection#close`.
+  # If a block is used the connection is yielded and closed automatically.
+  def self.connect(uri : URI | String)
+    build_connection(uri)
+  end
+
+  # Same as `#connect` but the database is yielded and closed automatically.
```
luislavena: Shouldn't this comment be `ditto`? Block behavior is already been mentioned in `self.connect` documentation. :smile:

PR #48, opened 2017-04-06
Shared specs

bcardiff:
This adds a helper class to ensure behaviour of custom drivers.

It is already been used in the three drivers and push to fix some issues in mysql and sqlite.
It also shorten the specs needed on each driver.

* https://github.com/crystal-lang/crystal-mysql/compare/shared-specs?expand=1
* https://github.com/crystal-lang/crystal-sqlite3/compare/shared-specs?expand=1
* https://github.com/will/crystal-pg/compare/master...bcardiff:shared-specs?expand=1

cc: @spalladino @asterite

### DIFF
--- a/src/spec.cr
+++ b/src/spec.cr
@@ -7,8 +7,48 @@ private def assert_single_read(rs, value_type, value)
 end
 
 module DB
+  # Helper class to ensure behaviour of custom drivers
+  #
+  # ```
+  # require "db/spec"
+  #
+  # DB::DriverSpecs(DB::Any).run do
+  #   # How to connect to database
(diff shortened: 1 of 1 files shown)


spalladino:
LGTM! I'd move the actual tests to a separate file than the helpers, just for the sake of clarity. But besides that, looks good.

bcardiff:
Well, the helper and the specs that it runs are pretty related. In order to avoid some unions and mimic real use cases the `#sample_value` needs to generate the specs for the values in that moment. So those need to be there. Splitting the specs in two parts does not seems to add any value.

spalladino:
Ok, let's keep it as a single file then

PR #50, opened 2017-04-21
WeakRef#target is now WeakRef#value

fridgerator:


bcardiff:
Thanks! I was waiting for 0.22 to be published :-)

Issue #55, opened 2017-06-20
query with as: NamedTuple


Nephos:
Hello,

I was wondering if you would want to implement something that could handle NamedTuple instead of just Tuples for queries ?

```crystal
DB.query_all("SELECT title, data, description FROM records ORDER BY created_at DESC LIMIT 10;", as: {String, String, String}).map { |r| {title: r[0], data: r[1], description: r[2]} }
# could be
DB.query_all("SELECT title, data, description FROM records ORDER BY created_at DESC LIMIT 10;", as: {title: String, data: String, description: String})
```

edit: I wrote the corresponding code and I open a PR in few minutes

edit: cf #56

ansarizafar:
Its a very convenient feature. Any update on this?

PR #56, opened 2017-06-21
Add NamedTuple argument for quering on the DB

Nephos:
It is now possible to write

```crystal
# old way
DB.query_all("SELECT title, data FROM records;", as: {String, String}).
  map { |r| {title: r[0], custom_name: r[1] }

# new way
DB.query_all("SELECT title, data FROM records;", as: {title: String, custom_name: String})
```

Let me now if you want me to add unit test / ...
I'm not 100% sure of the details on my own work.

#55

### DIFF
--- a/spec/dummy_driver_spec.cr
+++ b/spec/dummy_driver_spec.cr
@@ -138,7 +138,7 @@ describe DummyDriver do
         end
       end
 
-      it "with as named tuple" do
+      it "with a named tuple" do
         with_dummy do |db|
           db.query_one("3,4", as: {a: Int64, b: Int64}).should eq({a: 3i64, b: 4i64})
         end
@@ -182,6 +182,14 @@ describe DummyDriver do
         end
       end
 
+      it "with as" do
+        with_dummy do |db|
+          value = db.query_one?("3,4", as: {a: Int64, b: Int64})
+          value.should be_a(NamedTuple(a: Int64, b: Int64)?)
+          value.should eq({a: 3i64, b: 4i64})
+        end
+      end
+
       it "with as, just one" do
         with_dummy do |db|
           value = db.query_one?("3", as: Int64)
@@ -206,7 +214,7 @@ describe DummyDriver do
         end
       end
 
-      it "queries with as named tuple" do
+      it "queries with a named tuple" do
         with_dummy do |db|
           ary = db.query_all "3,4 1,2", as: {a: Int64, b: Int64}
           ary.should eq([{a: 3, b: 4}, {a: 1, b: 2}])
--- a/spec/dummy_driver_spec.cr
+++ b/spec/dummy_driver_spec.cr
@@ -138,6 +138,12 @@ describe DummyDriver do
         end
       end
 
+      it "with as named tuple" do
+        with_dummy do |db|
+          db.query_one("3,4", as: {a: Int64, b: Int64}).should eq({a: 3i64, b: 4i64})
+        end
+      end
+
       it "with as, just one" do
         with_dummy do |db|
           db.query_one("3", as: Int64).should eq(3i64)
@@ -197,6 +203,13 @@ describe DummyDriver do
         with_dummy do |db|
           ary = db.query_all "3,4 1,2", as: {Int64, Int64}
           ary.should eq([{3, 4}, {1, 2}])
+        end
+      end
+
+      it "queries with as named tuple" do
+        with_dummy do |db|
+          ary = db.query_all "3,4 1,2", as: {a: Int64, b: Int64}
+          ary.should eq([{a: 3, b: 4}, {a: 1, b: 2}])
         end
       end
 
--- a/src/db/query_methods.cr
+++ b/src/db/query_methods.cr
@@ -88,6 +88,20 @@ module DB
       end
     end
 
+    # Executes a *query* that expects a single row and returns it
+    # as a named tuple of the given *types*.
+    #
+    # Raises `DB::Error` if there were no rows, or if there were more than one row.
+    #
+    # ```
+    # db.query_one "select name, age from contacts where id = ?", 1, as: {name, String, age : Int32}
+    # ```
+    def query_one(query, *args, as types : NamedTuple)
+      query_one(query, *args) do |rs|
+        rs.read(**types)
+      end
+    end
+
     # Executes a *query* that expects a single row
     # and returns the first column's value as the given *type*.
     #
@@ -121,6 +135,23 @@ module DB
         value = yield rs
         raise DB::Error.new("more than one row") if rs.move_next
         return value
+      end
+    end
+
+    # Executes a *query* that expects a single row and returns it
+    # as a named tuple of the given *types*.
+    #
+    # Returns `nil` if there were no rows.
+    #
+    # Raises `DB::Error` if there were more than one row.
+    #
+    # ```
+    # result = db.query_one? "select name, age from contacts where id = ?", 1, as: {name: String, age: Int32}
+    # typeof(result) # => Tuple(String, Int32) | Nil
+    # ```
+    def query_one?(query, *args, as types : Tuple)
+      query_one?(query, *args) do |rs|
+        rs.read(*types)
       end
     end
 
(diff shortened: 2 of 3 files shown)


Nephos:
hello @bcardiff, some news ?

src/db/result_set.cr
```diff
@@ -99,6 +104,16 @@ module DB
       {% end %}
     end
 
+    private def internal_read(**types : **T) forall T
+      {% begin %}
+        NamedTuple.new(
+          {% for name, type in T %}
+            {{ name }}: read({{type.instance}}),
```
bcardiff: This won't work in the general case. The query determines the order of the columns, and the code expects it to be the order to the fields in the tuple. The order in the named tuples is not guaranteed since it can be interchanged without affecting the semantic.

Nephos: I see the point, however the only way I see to change that is to create tuple of named tuple or create a new structure

RX14: I kind of like that the naming is defined by the order not the names. Sometimes your DB will not use userscore_case identifiers and you want to convert them.

asterite: Order of named tuples is guaranteed. It's only when you mix two named tuples with same names and types but different orders that they get unified to one of them.

Nephos: Great.
@bcardiff is there another problem ?
(about the naming, there is no link between the name of the column and the name in the tuple, it is redondant in the current state of the PR)

asterite: Oh, but if you are passing a named tuple, then probably you should read values independent of the order, similar to how `DB.mapping` does it. You probably need to declare one variable for each key in the tuple, read them all, and construct the named tuple from all of that.

bcardiff: @Nephos what @asterite mention in the above message is the concern I raised in the initial review message. You can to be sure that the `read` is reading the column named `name` otherwise the result will be wrong. `read` read the column in the order of the query and that does not match the order of the names of the named tuple members.

ansarizafar:
This change would be very useful. Any update?

src/db/result_set.cr
Nephos: It does not need to be matched. I might want to use a shorter name than the column: "article.title" => :title

bcardiff: @Nephos The column has a fixed named, if `article.title` is used in sql query the name will be `title` AFAIK. I would expect that the name of the column should match exactly the name of the tuple entry.

RX14: And I disagree, as I said above. It's very common that the names in the DB are bad, or unidiomatic for crystal, and you cannot change them.

bcardiff: Maybe the best that `map_keys` can get is https://gist.github.com/bcardiff/b6c8fff62a5812e476a57e90c21a51c2 because you need some typing on that call, it needs to be a macro to be type safety.

Mapping keys it's something that would allow broader use cases for returning named tuples from db. I see them as independent topics.

@RX14 I'm sorry, but I don't get with what you disagree. Would you prefer a runtime error if the order of the query does not match the order of the keys in the named tuple?

RX14: I'd prefer it to entirely disregard the keys and read off the values in order which is exactly what this pr does now and is clearly workable because non-named triples do it.

Nephos: @bcardiff is that ok for you ?
(The name of the key is not associated with the column's names at all)
If not, we should close this PR, and I'll eventually make a fork (and synchronize it with this repo) to add this feature.

RX14: Perhaps we should have 2 names for the 2 behaviours.

Also, please can we move to traditional top-level comments from a line comment. They're more annoying and this change covers the whole scope of the PR.

src/db/query_methods.cr
```diff
@@ -88,6 +88,20 @@ module DB
       end
     end
 
+    # Executes a *query* that expects a single row and returns it
+    # as a named tuple of the given *types*.
+    #
+    # Raises `DB::Error` if there were no rows, or if there were more than one row.
+    #
+    # ```
+    # db.query_one "select name, age from contacts where id = ?", 1, as: {name, String, age : Int32}
```
bcardiff: I noticed a typo `as: {name, String, age : Int32}` -> `as: {name : String, age : Int32}`

src/db/result_set.cr
bcardiff: @Nephos yes, lets keep things simpler. The feature as it is is fine: the name of the key is not associated with the column's names at all. It's just a syntax sugar to avoid reading things in order and making a named tuple. 

If we could get some specs to cover the overloads that would be great.

src/db/query_methods.cr
Nephos: fixed

src/db/result_set.cr
Nephos: I'm on it

Nephos: Ok I've no idea where/how to make the specs :)

RX14: Here are the specs for the other `query_one` overloads: https://github.com/crystal-lang/crystal-db/blob/39e17f82ca1ce15145f68f0c21e9d1b00cb2f21b/spec/dummy_driver_spec.cr#L112

Just copy one for tuples, adapt it for named tuples.

Nephos: Ok thanks. I think it's ok.

spec/dummy_driver_spec.cr
```diff
@@ -138,6 +138,12 @@ describe DummyDriver do
         end
       end
 
+      it "with as named tuple" do
```
RX14: "with a named tuple" is enough, the `as` is implicit.

spec/dummy_driver_spec.cr
```diff
@@ -200,6 +206,13 @@ describe DummyDriver do
         end
       end
 
+      it "queries with as named tuple" do
```
RX14: ditto

spec/dummy_driver_spec.cr
Nephos: ok

spec/dummy_driver_spec.cr
Nephos: ok

Nephos:
I also added query_one(q, *a, as : NamedTuple)

src/db/query_methods.cr
```diff
@@ -124,6 +139,23 @@ module DB
...
+    #
+    # Returns `nil` if there were no rows.
+    #
+    # Raises `DB::Error` if there were more than one row.
+    #
+    # ```
+    # result = db.query_one? "select name, age from contacts where id = ?", 1, as: {name: String, age: Int32}
+    # typeof(result) # => Tuple(String, Int32) | Nil
+    # ```
+    def query_one?(query, *args, as types : Tuple)
```
RX14: This definition duplicates the definition directly below it. Perhaps you meant this to be a `NamedTpule` overload? Either way it should be deleted.

Nephos: I noticed that duplicate, but I did not see it in my commit

Nephos: Yes I wrote it 6 months ago. It was meant to be the NamedTuple overload, and has been lost with the time :D

src/db/query_methods.cr
```diff
@@ -125,21 +140,38 @@ module DB
     end
 
     # Executes a *query* that expects a single row and returns it
-    # as a tuple of the given *types*.
+    # as a named tuple of the given *types*.
```
RX14: Looks like you deleted the wrong one: the commit diff has doc changes now.

src/db/query_methods.cr
```diff
@@ -125,21 +140,38 @@ module DB
...
-    # result = db.query_one? "select name, age from contacts where id = ?", 1, as: {String, Int32}
+    # result = db.query_one? "select name, age from contacts where id = ?", 1, as: {name: String, age: Int32}
     # typeof(result) # => Tuple(String, Int32) | Nil
     # ```
     def query_one?(query, *args, as types : Tuple)
       query_one?(query, *args) do |rs|
         rs.read(*types)
       end
     end
+    # Executes a *query* that expects a single row and returns it
```
RX14: missing blank line between these two defs

src/db/query_methods.cr
Nephos: I try one last time, then I should get some sleep :)

Nephos:
@bcardiff Is it good for you ?

bcardiff:
Yes @Nephos, I was doubting to merge this before or after support 0.24 actually. Today or tomorrow I should take care of db* + 0.24. Thanks

bcardiff:
Thanks @Nephos for this! I released in 0.4.4 for those who wants/need to sticky with crystal 0.23 and 0.5 for crystal 0.24.

I am sorry it took so long to be on the same page.

PR #65, opened 2017-07-27
Release connection when query cannot be built

crisward:
This was leaking connection when a query could not be build - ie if an invalid column name was used.
More info here - https://github.com/crystal-lang/crystal-mysql/issues/40

### DIFF
--- a/spec/database_spec.cr
+++ b/spec/database_spec.cr
@@ -149,6 +149,28 @@ describe DB::Database do
     end
   end
 
+  it "should return connection to the pool if prepared statement is unable to be built" do
+    connection = uninitialized DB::Connection
+    with_dummy "dummy://localhost:1027?initial_pool_size=1" do |db|
+      connection = DummyDriver::DummyConnection.connections.first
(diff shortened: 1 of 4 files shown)


bcardiff:
* there was some on missing scenario (unprepared statements).
* there was a tracking on the connection that should not be done (although it was harmless).
* specs were needed.

@asterite or @RX14 do you want to double check?

asterite:
@bcardiff So cool that you added tests for this :-)

crisward:
@bcardiff Thanks, I wouldn't of had a clue on writing tests for this. Looking forward to seeing it merged and deleting my fork.

bcardiff:
Thanks you @crisward for digging and finding the right section for the fix :-).
My apologies for the delay here.

crisward:
@bcardiff That's ok. The nice thing about shards is it's pretty easy to use forks. Everyone finds it difficult to put time in on open source, I'm currently working on an imap library, which at it's current pace should be production ready in maybe 10 - 20 years.

PR #70, opened 2017-11-02
Updated exceptions so all exceptions can be caught from DB::Error

exts:
I noticed a few exceptions were being derived directly from `Exception` making them uncatchable with `DB::Error` and I fixed that.


RX14:
Looks like we need the core team to get access to every crystal-lang org repo to merge this.

bcardiff:
Thanks @exts

PR #71, opened 2017-11-13
The TEXT can always be read back as a String

### LINKED ISSUE #49: Readme example misleading
this line: `db.exec "create table contacts (name string, age integer)"`

should be `db.exec "create table contacts (name TEXT, age INT)"`

Because of sqlite's loose typing, if you change example line: `db.exec "insert into contacts values (?, ?)", "1234", 30`

this line will fail: `puts "#{rs.read(String)} (#{rs.read(Int32)})"`

thusly:

```
cast from Int64 to String failed, at ....../lib/db/src/db/result_set.cr:79 (TypeCastError)
```

Anyway might be confusing to people copy pasting examples

hinrik:
If you store something like "1234" in a STRING field, SQLite will return
an Int64. Better to be more specific, for uses who play around with the
example code.

Resolves #49.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ require "sqlite3"
 
 DB.open "sqlite3:./file.db" do |db|
   # When using the pg driver, use $1, $2, etc. instead of ?
-  db.exec "create table contacts (name string, age integer)"
+  db.exec "create table contacts (name text, age integer)"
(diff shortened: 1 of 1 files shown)


bcardiff:
Thanks @hinrik. I apply the same changes to the other related places. That copy paste error have gone wild!

PR #80, opened 2018-02-04
Add JSON::Any to DB::TYPES

docelic:
Enables reading out jsonb data from Postgres directly.
(Value returned is of type JSON::Any.)

### DIFF
--- a/src/db.cr
+++ b/src/db.cr
@@ -1,4 +1,5 @@
 require "uri"
+require "json"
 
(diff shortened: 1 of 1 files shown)


bcardiff:
Db drivers are able to return data of other types than DB::Any. Or at least that’s the intended behavior.

What is expected is that drivers support all of DB::Any types in order to have a clear common feature set.

docelic:
Ok, thanks

PR #89, opened 2018-08-14
Documentation improvements

MatthiasWinkelmann:
I tried to improve the instructions somewhat, based on the issues I encountered while getting up to speed with crystal-db myself.

Feel free to edit/merge partially/reject outright whatever you find use(ful/less).

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 # crystal-db
 
-Common db api for crystal. You will need a specific driver to access a database.
+Common DB API for Crystal. You will need a specific driver to access a database.
 
 * [SQLite](https://github.com/crystal-lang/crystal-sqlite3)
 * [MySQL](https://github.com/crystal-lang/crystal-mysql)
@@ -10,12 +10,20 @@ Common db api for crystal. You will need a specific driver to access a database.
 
 ## Installation
 
-Add this to your application's `shard.yml`:
+Add `crystal-db` to your application's `shard.yml`:
 
 ```yaml
 dependencies:
   db:
     github: crystal-lang/crystal-db
+```
+
+Add the driver for your database, i. e. for Postgres:
+
+```yaml
+dependencies:
+  pg:
+    github: will/crystal-pg
 ```
 
 ## Documentation
@@ -24,8 +32,6 @@ dependencies:
 * [Crystal book](https://crystal-lang.org/docs/database/)
 
 ## Usage
-
-Since this is an abstract db api, it's usage is through a concrete database driver.
 
 ```crystal
 require "db"
@@ -59,7 +65,7 @@ end
 
 ## Roadmap
 
-Issues not yet addressed
+Issues not yet addressed:
 
 - [x] Support non prepared statements. [#25](https://github.com/crystal-lang/crystal-db/pull/25)
 - [x] Time data type. (implementation details depends on actual drivers)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -2,31 +2,37 @@
...
   db:
     github: crystal-lang/crystal-db
 ```
 
+Add the driver for your database, i. e. for Postgres:
+
+```yaml
+dependencies:
+  pg:
+    github: will/crystal-pg
```
bcardiff: given that the usage sample is with sqlite (which requires less setup), I would suggest using sqlite here also. And I think is more clear to have only one yaml section with all the content. Otherwise it seems like the `db:` shard was removed here.

`i. e.` -> `i.e.` or `ie.`

README.md
```diff
@@ -2,31 +2,37 @@
...
+```
+
 ## Documentation
 
 * [Latest API](http://crystal-lang.github.io/crystal-db/api/latest/)
 * [Crystal book](https://crystal-lang.org/docs/database/)
 
 ## Usage
 
-Since this is an abstract db api, it's usage is through a concrete database driver.
```
bcardiff: I guess this was to abstract. What about

> You need to require the specific driver to use also.

So the user is warned to do `require "sqlite3"`

src/db.cr
```diff
@@ -95,15 +99,17 @@ module DB
     @@drivers ||= {} of String => Driver.class
   end
 
-  # Opens a database using the specified *uri*.
-  # The scheme of the *uri* determines the driver to use.
-  # Returned database must be closed by `Database#close`.
+  # Creates a `Database` pool and opens initial connection(s) as specified in the connection *uri*. Use `DB#connect` to open a single connection.
```
bcardiff: I think we tend to respect ~80 chars lines

src/db/database.cr
```diff
@@ -4,21 +4,20 @@ require "weak_ref"
 module DB
   # Acts as an entry point for database access.
   # Connections are managed by a pool.
-  # The connection pool can be configured from URI parameters:
+  #
+  # ## Database URI
+  #
+  # Connection parameters are configured in a URI. The format is specified by the individual database drivers. See [the book](https://github.com/crystal-lang/crystal-book/tree/master/database#open-database) for examples.
```
bcardiff: I am not sure about linking the book as samples, there are other samples that assume specific driver is used. 
Yet, if a link is to be used, use https://crystal-lang.org/docs/ instead of the source

src/db/database.cr
```diff
@@ -4,21 +4,20 @@ require "weak_ref"
...
   #   - retry_attempts (default 1)
   #   - retry_delay (in seconds, default 1.0)
-  #
-  # When querying a database prepared statements are used by default.
-  # This can be changed from the `prepared_statements` URI parameter:
-  #
-  #   - prepared_statements = `true`|`false` (default `true`)
-  #
-  # It should be created from DB module. See `DB#open`.
+  #   - prepared_statements (bool, default true)
```
bcardiff: I prefer to mention `true` or `false` directly since the value is expressed in a query string.

README.md
straight-shoota: What about

> This shard only provides an abstract database API. In order to use it, a specific driver for the intended database has to be required as well:

PR #94, opened 2018-11-13
Add the Cassandra driver

kaukas:
Does the "in development" need to be mentioned anywhere?


bcardiff:
Thanks @kaukas for building another driver :-) 

It's probably better to put the notice about the stage of the shard in its own repo so it can evolve.

kaukas:
Do you think the [current Status section](https://github.com/kaukas/crystal-cassandra#status) is clear enough?

On a different topic, [I had to jump through hoops](https://github.com/kaukas/crystal-cassandra/blob/master/spec/cassandra/custom_dbapi.cr#L10) to get the test suite to pass. Do you think it would make sense to somehow structure the tests differently? E.g. optional transaction support, a pluggable way to read scalars with less expectations, etc.

PR #101, opened 2019-04-11
Fix BeginTransaction#transaction error in rollback

straight-shoota:
When the rescued exception was caused by a protocol error, the
connection is in a corrupted state and executing the rollback
command might fail due to a subsequent protocol error - which is
irrelevant because the source of the corruption is described by
the original exception which must not be overlayed by a secondary one.


bcardiff:
Great change @straight-shoota

Issue #77, opened 2018-01-26
Pool issues


ysbaddaden:
With default values for the pool:
1. DB can create unlimited connections to the database;
2. DB may only keep 1 idle connection;
3. DB will disconnect connections on release (and keep 1 connection alive).

Since the default is to checkout/query/release, we easily end up with *many* connect / disconnect whenever we stress an application a little bit.

Maybe I'm wrong, but releasing a connection only checks `max_idle_pool_size` and doesn't seem to verify whether there are pending checkouts, which means that the connection wouldn't become idle.

In fact, I don't understand why the connection should be closed immediately. Why not push it back into the pool, and have a reaper that will close connections to fall below `max_idle_pool_size` after an idle threshold of some sort? That would avoid countless (dis)connect while still reducing the number of actual idle connections when the system is no longer under stress.

ysbaddaden:
After some debugging I found out that `@total.size` grows **far larger** than `max_pool_size`, and is only limited to how many connections are available on the SQL server.

For example with `max_pool_size=15&max_pool_idle_size=15&retry_attemps=5`; I have 300 incoming parallel requests that issue at most 5 SQL requests within a `db.using_connection` block. Debugging `@total.size` within `can_increase_pool_size`, it will stay at 1 or 2 for some time, then suddenly jump to 152, 175 then I start having `DB::ConnectionRefused` exceptions because my SQL server won't accept more than 200 incoming connections.

ysbaddaden:
There is a race condition in `DB#checkout` and `DB#build_resource` when called many times concurrently from many fiber.

`#build_ressource` calls `@factory.call` which create & connect a TCPSocket and possibly handshake... which will trigger the event loop, but we only push to `@total` *after* the factory returns, leaving an open window for a race condition in `#checkout`.

There must either be a lock in `#checkout` or keeping `@total_size` and `@available_size` as atomic values that will be incremented/decremented, but incapable to increment larger than maximum values, thus only being capable to increment `@total.size` would call `#build_resource` safely.

ysbaddaden:
In the meantime, the issue is avoided with `initial_pool_size  = max_pool_size = max_idle_pool_size`.

RX14:
Obviously the first step is benchmarking this outside framework benchmarks without any http code.

ysbaddaden:
@jwoertink yes, that was my workaround.

jwoertink:
[...]

PR #104, opened 2019-06-15
Adds a flags for the DB Pool before creating a new connection

Xosmond:
Hello, i was encountering the issue detailed in #77. So I make a flag to wait for a connection to complete to the database before trying to create a new one to avoid overflowing the `max_pool_size`.

I have tested this solution doing this using a personal project using Kemal deploying the project to two EC2 instances on AWS on a tunned OS. Ignore the timeouts since I was using a simple instance type behind a ELB to test my solution:

 The test:

Captura de Pantalla 2019-06-15 a la(s) 16 08 09

The connections on the database:

Captura de Pantalla 2019-06-15 a la(s) 16 07 40

My configuration was:
```
initial_pool_size: 20
max_idle_pool_size: 40
max_pool_size: 40
```

So max I should have: 80 connections (the other few are from other applications)

I'm not sure how I can integrate tests in my PR, because creating thousand of threads to test this doesn't looks like a good approach.

### DIFF
--- a/src/db/pool.cr
+++ b/src/db/pool.cr
@@ -11,6 +11,7 @@ module DB
     # maximum amount of retry attempts to reconnect to the db. See `Pool#retry`.
     @retry_attempts : Int32
     @retry_delay : Float64
+    @lock_checkout : Bool
 
     def initialize(@initial_pool_size = 1, @max_pool_size = 0, @max_idle_pool_size = 1, @checkout_timeout = 5.0,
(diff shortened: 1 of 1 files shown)


Xosmond:
I just closed this since on a regular CPU it worked fine with high load, nevertheless on a High power CPU it still doesn't solve the problem, it helps but doesn't solve the solve as it should be.

PR #109, opened 2019-08-20
Rework pool lock logic to handle new connections context switched

### LINKED ISSUE #77: Pool issues

bcardiff:
Fixes #77 

I was able to make a spec to repro the issue in https://github.com/crystal-lang/crystal-mysql/commit/325493b1f52aaacf00e495464b27f61b5e3a0554#diff-f41a8330fda8a5ae668e7e17638aef40R40-R57

A couple of things I considered during this PR:

* go's sql package added a `SetConnMaxLifetime` that would be good to have in order to limit the timespan of the whole connection. It's a good policy.
* When releasing a connection and if the max idle connection was reached, we could choose where to drop one in the idle pool rather than the one that is been released in order to keep the younger one.
* The checkout_some might be redundant. Maybe is just enough to reuse the statement and allow any connection to be used. If we do this, the idle connections could be kept in a Channel, as the message that is sent. But this will also require for the buffered channel to have a (theoretical) unlimited max capacity. 

For now, this PR should fix the issue but there are some ideas to improve the logic.


Xosmond:
@bcardiff Hey, great to know you worked on this PR, I will try it tomorrow to see it under pressure

src/db/pool.cr
```diff
@@ -2,47 +2,73 @@ require "weak_ref"
...
-    @initial_pool_size : Int32
-    # maximum amount of objects in the pool. Either available or in use.
-    @max_pool_size : Int32
-    @available = Set(T).new
-    @total = [] of T
-    @checkout_timeout : Float64
-    # maximum amount of retry attempts to reconnect to the db. See `Pool#retry`.
-    @retry_attempts : Int32
-    @retry_delay : Float64
+    # Pool configuration
```
straight-shoota: It would be better to put comments describing ivar's in a separate line before them, like we do with other features like method.
Even if they're not documented, it's preferable to use as single style.

bcardiff: done

src/db/pool.cr
```diff
@@ -2,47 +2,88 @@ require "weak_ref"
...
     def close : Nil
       @total.each &.close
       @total.clear
-      @available.clear
+      @idle.clear
+    end
+
+    # Returns stats of the pool
+    def stats
+      {open_connections: @total.size}
```
wontruefree: Are there any other stats that should be returned?
If not should there just be a method called `open_connections`?

bcardiff: I expect to add some but in a future PR. I needed it to write the spec in crystal-mysql

wontruefree: what else would be added?

bcardiff: https://github.com/golang/go/blob/master/src/database/sql/sql.go#L990-L1026 🙈

straight-shoota: This should better be a struct.

wontruefree: isnt a named tuple a fancy read only struct?

bcardiff: Yes. They are a struct. The GC.stats in the std-lib returns a record/struct. The difference is how to access the information. `s.prop` vs `s[:prop]`, how one would use `responds_to?`/`has_key?` and the ability to merge or shrink information in the later.

@straight-shoota is there any arguments why it should be better a struct? I like named tuples   over records usually for giving raw information.

straight-shoota: If nothing else, a struct can be easily documented what each property means. And I agree with what IIRC @asterite said, that the main use case for named tuples is named arguments and for every thing else it's probably better to use a defined data type.

bcardiff: There, the stats are now a `record`. Rebased and ready to merge upon 👍

src/db/pool.cr
```diff
@@ -38,20 +62,28 @@ module DB
...
-                     build_resource
+      res = sync do
+        resource = if @idle.empty?
+                     if can_increase_pool?
+                       @inflight += 1
+                       r = unsync { build_resource }
+                       @inflight -= 1
+                       r
+                     else
+                       unsync { wait_for_available }
```
RX14: By the time `unsync` has re-acquired the lock, there may be no idle connections to pick. Why not use a buffered channel as the idle connection pool itself?

RX14: Never mind, that was the next commit in the PR... :)

PR #110, opened 2019-09-05
Statement#exec and #query require named argument for array values

### LINKED ISSUE #100: Can't pass a single array as query param
The following query fails (using pg driver):
```cr
DB.open(ENV["DATABASE_URL"]) do |db|
  db.exec("SELECT $1::int[]", [1, 2])
end
```

Error message:
```
bind message supplies 2 parameters, but prepared statement "" requires 1
```

The call is essentially equivalent to `db.build("SELECT $1::int[]").exec([1, 2])`, so `QueryMethods#exec` builds a statement and forwards `args` to `Statement#exec`.

The reason for this fails is that `Statement#exec(*args)` is shadowed by `#exec(args : Array)` which can be used to pass a dynamic array of param bindings instead of varargs.
When the only query param is an array, this is interpreted as an array of params, instead of an array as single param.

A workaround is to wrap the array in another array: `db.exec("SELECT $1::int[]", [[1, 2]])` which will effectively call `#exec(args : Array)` with a params array including the intended array as single value.

While the workaround is relatively simple, this behaviour is very unexpected. Every other type works fine as a param provided as vararg. Array should just work as well.

A solution would be to make `args : Array` a required named argument: `#exec(*, args : Array)`. Thus, when calling `exec([1, 2])`, it passes the array `[1, 2]` as single param. When calling `exec(args: [1, 2])`, it passes two params, `1` and `2`. The delegating methods need to get support for this named argument as well.

(The prepared statement being empty in the error message is another issue).

straight-shoota:
This change allows to use an array as single argument for #exec and
 #query methods. Before it was shadowed by the *args splat overload.

Fixes #100

### DIFF
--- a/spec/custom_drivers_types_spec.cr
+++ b/spec/custom_drivers_types_spec.cr
@@ -227,14 +227,14 @@ describe DB do
       db.query "query", 1, "string" { }
       db.query("query", Bytes.new(4)) { }
       db.query("query", 1, "string", FooValue.new(5)) { }
-      db.query "query", [1, "string", FooValue.new(5)] { }
+      db.query "query", args: [1, "string", FooValue.new(5)] { }
 
       db.query("query").close
       db.query("query", 1).close
       db.query("query", 1, "string").close
       db.query("query", Bytes.new(4)).close
       db.query("query", 1, "string", FooValue.new(5)).close
-      db.query("query", [1, "string", FooValue.new(5)]).close
+      db.query("query", args: [1, "string", FooValue.new(5)]).close
     end
 
     DB.open("bar://host") do |db|
@@ -244,14 +244,14 @@ describe DB do
       db.query "query", 1, "string" { }
       db.query("query", Bytes.new(4)) { }
       db.query("query", 1, "string", BarValue.new(5)) { }
-      db.query "query", [1, "string", BarValue.new(5)] { }
+      db.query "query", args: [1, "string", BarValue.new(5)] { }
 
       db.query("query").close
       db.query("query", 1).close
       db.query("query", 1, "string").close
       db.query("query", Bytes.new(4)).close
       db.query("query", 1, "string", BarValue.new(5)).close
-      db.query("query", [1, "string", BarValue.new(5)]).close
+      db.query("query", args: [1, "string", BarValue.new(5)]).close
     end
   end
 
@@ -263,7 +263,7 @@ describe DB do
       db.exec("query", 1, "string")
       db.exec("query", Bytes.new(4))
       db.exec("query", 1, "string", FooValue.new(5))
-      db.exec("query", [1, "string", FooValue.new(5)])
+      db.exec("query", args: [1, "string", FooValue.new(5)])
     end
 
     DB.open("bar://host") do |db|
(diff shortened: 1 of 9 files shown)


bcardiff:
@straight-shoota what would you think of https://github.com/crystal-lang/crystal-db/commit/f11ae18c6e5d10625326c37afa7963205103c75e ?

straight-shoota:
Hm, I'm not sure... Avoiding duplication is good. Really good. But now the signatures are a bit cryptic. We could probably find a better name than `t_args`. But it still suggests you could use them in combination with `args`. It won't compile, so it's not that big of an issue from the implementation side. But I'm thinking of the API documentation here. More explicitness might be favourable. OTOH it means so many method with subtle differences that you're just overwhelmed.

bcardiff:
If the overload approach is used, is harder to check if all the overloads are implemented. `StatementMethods#scalar(args:)` and `StatementMethods#query(args:,&block)` were missing for example.

A runtime check to ensure `t_args` is empty if `args` is not nil could be added. But is a runtime check.

Using some metaprogramming here seems overkill. So is either:

* shorter code easier to maintain, easier to see all the alternatives, a bit criptic signatures, but can be clarified, allow `query("stmt", 1,2, args: [1, 2])` to compile. 

Or,

* more code, chances of not covering all the cases, better documentation for each case, but harder to see it as a whole.

straight-shoota:
As a slick alternative, when given both positional args and `args`, they could just be concatenated. So `query(1, 2, args: [3, 4])` would pass `[1, 2, 3, 4]` as arguments. I don't think there is an important use case for this, but it's a valid solution and would remove the weirdness of choosing either one and potentially producing a runtime error.

Thinking about it, this might actually prove somewhat useful for combining static with dynamic arguments.

Btw, your proposed implementation surely doesn't work with `t_args || args` because `t_args` can't be falsey.

bcardiff:
I rebased on master and pushed here for discussion.
I made the way to merge all the args without incurring in additional allocations.
I'm satisfied with the result so far. Are you @straight-shoota?

straight-shoota:
Yes, it looks good.
I still have a few thought though:

I'd rename `t_args` to a more descriptive name, maybe `positional_args`.

Do we really need `EnumerableConcat`? I thought the concatenation could just create a new array and append all arguments. This requires a heap allocation and memcopy, but only when both argument types are used (which should be the exception), if you want more performance, you can pass in all arguments in a single collection.
Alternatively, we could consider adding `EnumerableConcat` as a general purpose type to the stdlib. But just having it for this rare case of concatenating positional and array arguments, seems a bit heavy.

bcardiff:
I don’t like positional arguments because I think is missleading to the argumenta of the query from the driver perspective. We could use `args_` or something alike.

The EnumerableConcat is a struct. There is no heap allocation. My goal was to avoid heap allocation for executing a command. The drivers assumes an enumerable that is actually a tuple or an array before this PR.

straight-shoota:
Hm, yeah okay. `args_` would probably be better than `t_args` ("what does t mean?"). Or maybe expand it to `tuple_args`?

I meant that concatenating into a new array would require an allocation. But it's probably not a big deal, because you'd usually either use tuple args or an enumerable, not a combination. IMO it's okay that using a combination requires allocation.

straight-shoota:
> The drivers assumes an enumerable that is actually a tuple or an array 

Maybe this should be a separate issue then. Technically, it should be fine to provide *any* `Enumerable`, not just `Array` or `Tuple`.

bcardiff:
For now dynamic argument list must be an Array. We can relax that later if needed.
Something that I also have in mind is if there should be support for named argument.

straight-shoota:
:+1: (can't approve my own PR)

PR #115, opened 2019-10-27
Add `DB::Serializable`

### LINKED ISSUE #105: Add DB::Serializable (à la JSON::Serializable)
I just started adding database mappings to an API that I'm writing, but I'm having trouble keeping my schema definitions concise. I currently have everything already using the `JSON::Serializable` module, as I think it's more effective than the equivalent `mapping` macro. However, when adding database support, this adds a weird disconnect:
```crystal
class Tag
  include JSON::Serializable

  DB.mapping({
    id: Int32,
    name: String,
    count: Int32,
    type: String
  })
end
```
This works fine as-is, but I strongly dislike having to mix both systems in the same class and I would much rather be able to do something like this:
```crystal
class Tag
  include JSON::Serializable
  include DB::Serializable

  property id : Int32
  property name : String
  property count : Int32
  property type : String
end

### PR DESCRIPTION
**Pull request by nickbclifford**
This PR adds the `DB::Serializable` module and corresponding `DB::Field` annotation in analogue with stdlib `JSON::Serializable`. Resolves #105.

### CODE REVIEW

**Comment by pynixwang** [comment]
`Model.from_rs` should return only one instance.
`Array(Model).from_rs` return an array is better

**Comment by RX14** [comment]
@bcardiff I agree with @pynixwang here, what do you think about making this change?

**Comment by bcardiff** [comment]
It bugged me in the past the distinction between `Model.new(rs)` and `Model.from_rs(rs)`. So yeah, we can change the design.

There was no decision in this PR, it was based on the [current desing of DB::Mapping](https://github.com/crystal-lang/crystal-db/blob/7e83f8e26c2f9a5cde0693007b1c1486d00f7685/src/db/mapping.cr#L88-L96).

A nice thing to achieve/keep is that the `db.query_all|one|each("...", as: Model)` work yielding or returning the appropriate type. I thing query_each might be missing. I think those are the main API and `from_rs` is more an internal thing.

@pynixwang what pushed you to use directly `from_rs`.

**Comment by pynixwang** [comment]
@bcardiff  use like `JSON`.

or `form_db`?

**Comment by bcardiff** [comment]
Can I get more details? :-)

`query_all` is already something that returns Array. So I see little benefit in having an internal like `from_rs` to do that also.

To iterate each entity directly you could either avoid having a rs at all by using `db.query_...("...", as: Model)`. Or at least `rs.read(Model)` thanks to [this definition](http://crystal-lang.github.io/crystal-db/api/0.8.0/DB/ResultSet.html#read(type:DB::Mappable.class)-instance-method). (`DB::Serializable` includes `DB::Mappable`).

So I see no need to explicitly call `from_rs`. Moving this conversation to an internal design discussion (which are easier to act upon if they don't change the main API).

**Comment by pynixwang** [comment]
ok, I see.
not using it in project cause a mistake.

**Comment by pynixwang** [comment]
@bcardiff the doc lead me to a black hole

http://crystal-lang.github.io/crystal-db/api/0.8.0/DB/Serializable.html

```
employees = Employee.from_rs(db.query("SELECT title, name FROM employees"))
```

pynixwang:
this doc must be changed

PR #121, opened 2020-02-08
making exception error string a proper error class.

### LINKED ISSUE #120: A better error for "no results"
Right now if your query doesn't return anything, an exception is raised saying [no results](https://github.com/crystal-lang/crystal-db/blob/master/src/db/statement.cr#L18).

I'd like to propose this become a special class like `raise NoResultsException.new("no results")`. This would allow libraries that wrap this to catch on that error specifically to handle things specialized for that library.

jwoertink:
Fixes #120

I was going to write a spec for it, but the specs in this are sort of complicated, and I wasn't sure how to test that the scalar raises an exception.

### DIFF
--- a/spec/statement_spec.cr
+++ b/spec/statement_spec.cr
@@ -208,4 +208,13 @@ describe DB::Statement do
       db.pool.is_available?(DummyDriver::DummyConnection.connections.first)
     end
   end
+
+  it "raises NoResultsError for scalar" do
+    with_dummy_connection do |cnn|
+      stmt = cnn.prepared ""
+      expect_raises DB::NoResultsError do
+        stmt.scalar "SELECT LIMIT 0"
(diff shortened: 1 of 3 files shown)


straight-shoota:
How about `SELECT LIMIT 0`?

jwoertink:
I tried doing this:

```crystal
    with_dummy do |db|
      expect_raises DB::NoResultsError do
        db.scalar "SELECT LIMIT 0"
      end
    end
```

but it never raises an exception.

jwoertink:
ah, nevermind, I was using the wrong thing.

jwoertink:
Hey, just wanted to ping this. If anyone has a second to look over it. Thanks!

straight-shoota:
Thanks @jwoertink

PR #125, opened 2020-04-06
Use DB::NoResultsError on QueryMethods

bcardiff:
Follow up of #121 since there were other places to raise that exception

cc: @jwoertink


jwoertink:
Ah! Good catch. Yeah, that will be a lot nicer to work with.

PR #128, opened 2020-04-07
Fix mutex deadlock in setup_connection

### LINKED ISSUE #127: Mutex deadlock in setup_connection
I'm using the `setup_connection` hook to configure newly established connections on crystal-pg.
```cr
db.setup_connection do |connection|
  if application_name = self.application_name
    connection.exec "SET application_name TO '#{application_name.dump_unquoted}';"
  end
  connection
end
```
This ends up triggering a mutex deadlock:
```
Unhandled exception: Attempt to lock a mutex recursively (deadlock) (Exception)
         from /usr/share/crystal/src/mutex.cr:44:21 in 'lock'
         from /src/lib/db/src/db/pool.cr:242:7 in 'release'
         from /src/lib/db/src/db/database.cr:85:7 in 'release'
         from /src/lib/db/src/db/connection.cr:78:7 in 'release'
         from /src/lib/db/src/db/connection.cr:83:7 in 'release_from_statement'
         from /src/lib/db/src/db/statement.cr:58:7 in 'release_connection'
         from /src/lib/db/src/db/statement.cr:84:7 in 'perform_exec_and_release'
         from /src/lib/db/src/db/statement.cr:68:7 in 'exec:args'
         from /src/lib/db/src/db/query_methods.cr:271:7 in 'exec'
         from /src/lib/shardbox-core/src/db.cr:24:11 in '->'
         from /src/lib/db/src/db/database.cr:255:3 in 'setup_connection'
         from /src/lib/shardbox-core/src/db.cr:22:7 in 'db'
         from /src/lib/shardbox-core/src/service/worker_loop.cr:112:20 in 'perform'
         from /src/lib/shardbox-core/src/worker.cr:66:3 in '__crystal_main'
         from /usr/share/crystal/src/crystal/main.cr:105:5 in 'main_user_code'
         from /usr/share/crystal/src/crystal/main.cr:91:7 in 'main'
         from /usr/share/crystal/src/crystal/main.cr:114:3 in 'main'
         from ???
```

I can't pinpoint the exact reasons, but disabling the setup_connection hook makes the error go away.
For some reason it only reprodues in production. I can't really trigger the error in any other way, even from the exact same docker image. So, something seems really weird, and I'm not sure if crystal-db is at fault here. But it might be.

straight-shoota:
Adds auto_release = false in setup_connection to avoid trying to release
the connection to the pool before it has been added.

Fixes #127

### DIFF
--- a/spec/database_spec.cr
+++ b/spec/database_spec.cr
@@ -8,6 +8,7 @@ describe DB::Database do
 
       db.setup_connection do |cnn|
         cnn_setup += 1
+        cnn.scalar("1").should eq "1"
       end
 
       cnn_setup.should eq(2)
--- a/src/db/database.cr
+++ b/src/db/database.cr
@@ -55,6 +55,7 @@ module DB
       @pool = uninitialized Pool(Connection) # in order to use self in the factory proc
       @pool = Pool.new(**pool_options) {
         conn = @driver.build_connection(self).as(Connection)
+        conn.auto_release = false
         @setup_connection.call conn
         conn
       }


straight-shoota:
I'm not sure if auto_release should be set to `true` later, but doing it after `@setup_connection.call conn` still failed the spec and without it, everything's green...

straight-shoota:
@bcardiff Can you take a look at this?

PR #129, opened 2020-05-11
Improve DB::MappingException usage

straight-shoota:
This adds more locations where DB::MappingException is raised and it adds information about the type and property that failed to deserialize.

cf. https://github.com/crystal-lang/crystal/pull/9274

### DIFF
--- a/src/db/error.cr
+++ b/src/db/error.cr
@@ -3,6 +3,19 @@ module DB
   end
 
   class MappingException < Error
+    getter klass
+    getter property
+
+    def initialize(message, @klass : String, @property : String? = nil, cause : Exception? = nil)
+      message = String.build do |io|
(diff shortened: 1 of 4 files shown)


src/db/error.cr
```diff
@@ -3,6 +3,19 @@ module DB
   end
 
   class MappingException < Error
+    getter klass
```
bcardiff, line 6: Is there a reason why you prefer the to store the class name instead of the metaclass directly?

straight-shoota: IMO it's good to keep exception types as simple es possible. Having the metaclass directly doesn't add a real benefit. It's just informational and a string should suffice. This makes it easier to introspect/export an exception instance.

PR #131, opened 2020-06-12
Allow DB::Pool to be used a generic connection pool

### LINKED ISSUE #122: Extract ConnectionPool to its own shard?
I've been using the [`pool` shard](https://github.com/ysbaddaden/pool) for a generic connection pool but I had some issues with it when using `-Dpreview_mt` and, since the `ConnectionPool` implementation in this shard is both powerful and threadsafe, I wonder if it would be useful to extract it to its own shard to become the de facto standard connection pool implementation.

I've used it in the [`neo4j` shard](https://github.com/jgaskins/neo4j.cr/blob/add-cluster-support/src/neo4j/connection_pool.cr) and applied only minimal tweaks to get it working there — rather than add checkout callbacks to the connection, I removed their calls in the pool.

Any thoughts on this?

jgaskins:
This allows an application to use `DB::Pool` for all of its connection pools. I've included a couple examples:

- `HTTP::Client` — the idea is that you'd wrap a `Pool` instance to make your own HTTP API client
- a fake `IO` object which I was using to check that concurrency was still working

I'd be cool with removing the examples. They're just what I was using to make sure this was truly generic.

Fixes #122


src/db/pool.cr
```diff
@@ -91,10 +100,19 @@ module DB
...
+        res.before_checkout
+      end
       res
     end
 
+    def checkout(&block : T ->)
+      connection = checkout
+      yield connection
+    ensure
+      release connection if connection
```
bcardiff, line 113: why the `if`?

src/db/pool.cr
```diff
@@ -52,12 +54,19 @@ module DB
       @idle.clear
     end
 
-    record Stats, open_connections : Int32
+    record Stats,
```
bcardiff, line 57: Thanks for adding stuff in the stats.

src/db/pool.cr
```diff
@@ -91,10 +100,19 @@ module DB
...
       end
 
-      res.before_checkout
+      if res.responds_to?(:before_checkout)
+        res.before_checkout
+      end
       res
     end
 
+    def checkout(&block : T ->)
```
bcardiff, line 109: I'm not sure. But the new `Pool#checkout(&)` might not play nice with DB connections. It's not documented, but the `auto_release` property handles the logic of how the resource is released after the operation is completed. This requires some knowledge of the resource though. Maybe we need to formalise that a bit better.

examples/http_client.cr
```diff
@@ -0,0 +1,8 @@
+require "http"
```
bcardiff, line 1: I think we can create specs to ensure the pool is generic.

examples/http_client.cr
```diff
@@ -0,0 +1,8 @@
...
+require "../src/db/pool"
+
+pool = DB::Pool.new { HTTP::Client.new(URI.parse("https://google.com")) }
```
bcardiff, line 4: If we want `DB::Pool` to be used directly maybe the API for setting the pool options should be refactored a bit.

Otherwise it will be too fragile when new options are coming.

src/db/error.cr
```diff
@@ -14,10 +14,10 @@ module DB
   # Raised when an established connection is lost
   # probably due to socket/network issues.
   # It is used by the connection pool retry logic.
-  class ConnectionLost < Error
-    getter connection : Connection
+  class ConnectionLost(T) < Error
```
bcardiff, line 17: Note to future self: check if this is causing any breaking change on db drivers

src/db/pool.cr
```diff
@@ -215,7 +235,7 @@ module DB
           sync_dec_waiting_resource
         when timeout(@checkout_timeout.seconds)
           sync_dec_waiting_resource
-          raise DB::PoolTimeout.new
+          raise PoolTimeout.new("Could not check out a connection in #{@checkout_timeout} seconds")
```
bcardiff, line 238:
```suggestion
          raise DB::PoolTimeout.new("Could not check out a connection in #{@checkout_timeout} seconds")
```

for consistency 🙈

src/db/pool.cr
jgaskins: An exception could theoretically be raised in `checkout`, so there are no compile-time guarantees that `connection` is non-nil.

I could scope the `ensure` block only to the `yield` if we want to remove the `if`:

```crystal
def checkout(& : T ->)
  connection = checkout

  begin
    yield connection
  ensure
    release connection
  end
end
```

But I don't know if that's any clearer.

src/db/pool.cr
jgaskins: Interesting. I haven't looked into how that works but it sounds like it might fall under `DB`-specific extensions that could be made if this class were extracted to its own shard, which `DB::Pool` could then inherit from.

examples/http_client.cr
jgaskins: I had a feeling you might say that 🙂 I didn't know how to do it, though, so I figured I'd at least offer up what I had.

To clarify, I've always struggled with testing I/O. I don't know that we'd have to _actually_ test I/O here, but that hadn't crossed my mind when I was writing this.

src/db/error.cr
jgaskins: I've checked this against `pg` and it worked fine. I didn't check the other drivers, though.

src/db/pool.cr
bcardiff: the latter seems cleaner to me.

bcardiff:
@jgaskins I checked the status of this

- added a spec to ensure http client is usable with the pool in https://github.com/crystal-lang/crystal-db/commit/157660075950b8e411c1434b57b69f3d93feae07 (and deleted the examples)
- workaround an ICE related to https://github.com/crystal-lang/crystal/issues/8721 in https://github.com/crystal-lang/crystal-db/commit/d0050eb994fca30b1b5177aa2fa924352f8452a9

The pool still requires that the resource is a Disposable (or at least has a close method) something that the HTTP::Client has. So far if this status is useful to you, we can add those commits here and merge this PR.

jgaskins:
@bcardiff That sounds great. The specs look good! Nice catch on the ICE, too — but what does that stand for?

I was aware of the requirement that it respond to `close` and, honestly, I think it makes sense. Since this is intended to be used for connections I imagine most people using this will be managing sockets or some other `IO` with them. I use it [in my own Redis client](https://github.com/jgaskins/redis/blob/ede13b3741a4a0d92bf50a758e452f60de510831/src/client.cr#L20) for this purpose.

Are you able to push to that branch on my fork since I opened this PR or do you need me to pull them in?

bcardiff:
ICE = Internal Compiler Error.

Yup, I didn't want to push before asking.

jgaskins:
Awesome, thank you for your help with this. I really wasn't sure how to automate the tests for that and, looking at the ones you wrote here, I feel like it would've taken me a lot of time and experimentation to arrive at anything resembling a useful spec, so I really appreciate your effort there.

bcardiff:
Welcome. The support files are 99% the ones in the std-lib from the http server specs and fibers specs. It's hard to express exactly the expected interactions when dealing with fibers and ensuring the specs will not hang forever is something that requires extra attention.

I needed to add one more fix and that is effectively a breaking change, but required to allow the right type of exception to be raised and caught in the pool.

If nobody complains about that, we can merge this in a couple of days :-)

PR #134, opened 2020-09-14
Add logging for executing queries

### LINKED ISSUE #62: Query Logging
I realise it's possible to do this with access to the db. But it would be useful, especially when debugging performance to be able to get the raw queries logged. 

ie
```crystal
db = DB.open(ENV["DATABASE_URL"])
db.logger = ->mylogger(String)
db.logging = true
# do some stuff
db.logging = false

def mylogger(sql : String)
  puts "#{Time.new} #{sql}"
end
```

I could imagine using this to get all the queries used on a single request. Though I'm unsure if this would work with the async nature of how queries are handled. I'll happily give this a go if anyone else thinks this is feasible.

bcardiff:
This PR adds a debug "Executing query" message to the `db` log source. The query and the arguments are passed as local metadata `query` and `args.

The arguments are translated to `Log::Metadata::Value` via `DB::Statement#arg_to_log` method.

For example in crystal-pg the following code would allow customizing how the `Geo::Point` argument is translated.

```crystal
class DB::Statement
  private def arg_to_log(arg : PG::Geo::Point)
    Log::Metadata::Value.new("(#{arg.x}, #{arg.y})::point")
  end
end
```

By default, `arg.to_s` is used for all the values that are not directly supported by `Log::Metadata::Value`.

`DB::Statement#before_query_or_exec` & `after_query_or_exec` protected methods can be used to hook around the statement execution.

The `DB::Statement` has now a `command : String` property that is specified during initialization. This the breaking-change part of this PR.

Fixes #62
Closes #64

### DIFF
--- a/src/db/enumerable_concat.cr
+++ b/src/db/enumerable_concat.cr
@@ -20,7 +20,7 @@ module DB
     end
 
     # returns given `e1 : T` an `Enumerable(T')` and `e2 : U` an `Enumerable(U') | Nil`
-    # it returns and `Enumerable(T' | U')` that enumerates the elements of `e1`
+    # it returns an `Enumerable(T' | U')` that enumerates the elements of `e1`
     # and, later, the elements of `e2`.
     def self.build(e1 : T, e2 : U)
       return e1 if e2.nil? || e2.empty?
--- a/src/db/enumerable_concat.cr
+++ b/src/db/enumerable_concat.cr
@@ -20,7 +20,7 @@ module DB
     end
 
     # returns given `e1 : T` an `Enumerable(T')` and `e2 : U` an `Enumerable(U') | Nil`
-    # it retuns and `Enumerable(T' | U')` that enumerates the elements of `e1`
+    # it returns and `Enumerable(T' | U')` that enumerates the elements of `e1`
     # and, later, the elements of `e2`.
     def self.build(e1 : T, e2 : U)
       return e1 if e2.nil? || e2.empty?
--- a/spec/custom_drivers_types_spec.cr
+++ b/spec/custom_drivers_types_spec.cr
@@ -48,11 +48,11 @@ class FooDriver < DB::Driver
   end
 
   class FooConnection < DB::Connection
-    def build_prepared_statement(query) : DB::Statement
-      FooStatement.new(self)
+    def build_prepared_statement(command) : DB::Statement
+      FooStatement.new(self, command)
     end
 
-    def build_unprepared_statement(query) : DB::Statement
+    def build_unprepared_statement(command) : DB::Statement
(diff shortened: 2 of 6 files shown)


asterite:
This is great! But extending the way logging works by adding overloads to undocumented, private methods might be a bit confusing or obscure.

How about introducing an "interface" (a module) that provides a way to turn an argument into a DB log argument, make the existing objects implement it, and let the implementation use `to_s` for objects that don't conform to that interface?

Or some other way to make this a bit less hidden.

bcardiff:
I didn't want to pollute the Object with this. And is something only drivers will need.

Having a public module to encapsulate that interface is good enough for me. Does that sounds good to you?

jwoertink:
So this is something where crystal-pg, in this case, would implement any overrides with `arg_to_log` in the `PG::Statement` class it has? If ORMs wanted to add in their own, we'd have to create a new sub class or override these existing?

bcardiff:
@jwoertink overriding the method will do.

bcardiff:
With the last commit I applied the same pattern as in https://github.com/crystal-lang/crystal/pull/9756 .
It will allow measuring the elapsed time in the query without additional heap memory.

```crystal
class DB::Statement
  def_around_query_or_exec do |args|
    start = Time.monotonic
    res = yield
    elapsed = Time.monotonic - start
    Log.debug &.emit("Finished", elapsed: "#{elapsed.total_seconds.humanize(precision: 2, significant: false)}s")
    res
  end
end
```

src/db/enumerable_concat.cr
```diff
@@ -20,7 +20,7 @@ module DB
     end
 
     # returns given `e1 : T` an `Enumerable(T')` and `e2 : U` an `Enumerable(U') | Nil`
-    # it retuns and `Enumerable(T' | U')` that enumerates the elements of `e1`
+    # it returns and `Enumerable(T' | U')` that enumerates the elements of `e1`
```
asterite, line 23:
```suggestion
    # it returns an `Enumerable(T' | U')` that enumerates the elements of `e1`
```

asterite (approved):

bcardiff:
The `DB::MetadataValueConverter#arg_to_log` is the final location to define how arguments are translated to Log::Metadata::Value

```crystal
class DB::MetadataValueConverter
  def arg_to_log(arg : PG::Geo::Point)
    Log::Metadata::Value.new("(#{arg.x}, #{arg.y})::point")
  end
end
```

And the `DB::Statement.def_around_query_or_exec` macro as explained in https://github.com/crystal-lang/crystal-db/pull/134#issuecomment-695081824 is to be used to hook into the command execution lifecycle.

PR #142, opened 2021-01-20
Fix model.from_rs argument type typo

dukeraphaelng:
Fixed a typo of 

```crystal
def self.from_rs(rs : ::DB::Result_set)
```
->
```crystal
def self.from_rs(rs : ::DB::ResultSet)
```

- Resolves https://github.com/crystal-lang/crystal-db/issues/143

### DIFF
--- a/src/db/serializable.cr
+++ b/src/db/serializable.cr
@@ -95,7 +95,7 @@ module DB
           super
         end
 
-        def self.from_rs(rs : ::DB::Result_set)
+        def self.from_rs(rs : ::DB::ResultSet)
           super
         end
       end


jgaskins:
If this hasn't been causing any compiler errors in the past 2 years since it was added, it seems to imply the method isn't being used, right? If that's the case, should the method def instead be removed?

oprypin (approved):

straight-shoota (approved):

straight-shoota:
Maybe the method could be removed, but let's make that a separate issue.

PR #144, opened 2021-01-20
Containerise crystal-db for CI

dukeraphaelng:
CI is reporting that specs fail on the master branch even though the latest commit was only a change to comments/ documentation, and not to the code, (https://github.com/crystal-lang/crystal-db/commit/eaddae7d71d52536453ea2d94777854f1b057f83) and consecutively before that CI was passing.

This suggested that environmental differences (OS) compromise the reproducibility of the test. To counter this, I thought it would be a good idea to containerise the test build for CI so this situation would not occur again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,11 @@
 language: crystal
+
+services:
+  - docker
+
 crystal:
   - latest
   - nightly
-script:
-  - crystal spec
-  - crystal tool format --check
+
+  script:
+  - docker build -t crystal_db . && docker run crystal_db
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,5 +7,5 @@ crystal:
   - latest
   - nightly
 
-  script:
+script:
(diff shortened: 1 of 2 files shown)


straight-shoota:
I don't see how this changes anything. Travis job runners already execute in a standard image, so there's no variation on that part which this would help to mitigate.

In any case, I think we should probably consider migrating from travis to GitHub Actions as the main CI platform like in the main repo.

straight-shoota:
The [current failure on nightly](https://travis-ci.org/github/crystal-lang/crystal-db/jobs/755116757) is caused by https://github.com/crystal-lang/crystal/pull/10185#issuecomment-762467026 and already fixed in Crystal master. Today's nightly dist job just failed, again 😠 (https://github.com/crystal-lang/crystal/issues/10175), so the changes are not applied yet.

dukeraphaelng:
I see. Gotcha!

PR #146, opened 2021-01-31
Return nothing from transaction blocks

### LINKED ISSUE #145: Transactions don't return the result of the yielded block
It currently returns a boolean of whether or not the transaction was committed successfully https://github.com/crystal-lang/crystal-db/blob/eaddae7d71d52536453ea2d94777854f1b057f83/src/db/begin_transaction.cr#L29

This is different than Rails, for instance, that does return that value https://github.com/rails/rails/blob/d5386cfba2ec5c80f09e37d44b87bcae9c6ab6a8/activerecord/lib/active_record/connection_adapters/abstract/transaction.rb#L310-L312

I would rather get the result of the block rather than the vague boolean, though I know this will be a breaking change.

matthewmcgarvey:
Fixes #145

### DIFF
--- a/spec/transaction_spec.cr
+++ b/spec/transaction_spec.cr
@@ -175,4 +175,14 @@ describe DB::Transaction do
       db.pool.is_available?(cnn).should be_true
     end
   end
+
+  it "transaction with block returns result of block" do
+    with_dummy_connection do |cnn|
+      result = cnn.transaction do |_tx|
+        "RESULT"
+      end
+
+      result.should eq "RESULT"
+    end
+  end
 end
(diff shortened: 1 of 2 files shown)


bcardiff:
The resulting type will be nillable. Because of the DB::Rollback path. Will this be comfortable enough on the use case you have in mind?

matthewmcgarvey:
@bcardiff I believe so. We don't have any direct need for the return value but the pain point for Avram was the boolean return type.

straight-shoota:
Maybe the return type should just be `Nil`. Returning the result of the block or nil doesn't seem very great. If there's no real use case to return the actual block value, we could just make it nil.

matthewmcgarvey:
@straight-shoota I changed it to return Nil, but thinking about it now if that is better than returning the block result then this is really a non-issue

PR #155, opened 2021-09-06
feat(error): close resources on resource lost

stakach:
this allows exceptions to be handled outside of the pool logic without risk of the resource being returned to the pool

### DIFF
--- a/spec/database_spec.cr
+++ b/spec/database_spec.cr
@@ -172,6 +172,21 @@ describe DB::Database do
     end
   end
 
+  it "should close connection on ConnectionLost" do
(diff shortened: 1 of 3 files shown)


bcardiff:
Spec and note added :-)

stakach:
@straight-shoota might be worth you reviewing this one?
With this change Clear ORM in its current state will reconnect properly.

The feature is that ORMs / applications can handle this error and the connection won't be returned to the pool.
Separately we're working on adding the retry behaviour to Clear

straight-shoota (approved):

sdogruyol (approved):
Thank you @stakach 🙏

Issue #22, opened 2016-10-20
better error message when types don't line up


rdp:
Today I get this:

cast from Int32 to Int64 failed, at /x/y/libs/db/db/result_set.cr:80 (TypeCastError)

if I declare an "id" column as Int64 instead of Int32 that it is in the DB.

Might be nice to say which column is failing when this occurs instead of "somewhere, there is a failure in the conversion..."

asterite:
To be able to do this all ResultSets should store the current column index, so they can retrieve the current column name from that index.

rdp:
actually, just outputting the current column index would be an improvement :)

PR #156, opened 2021-10-03
Raise a specific class error instead of string literal

jwoertink:
Related: #22 

I don't think it quite solves the original issue since a better solution would be to show *which* column didn't match; however, by adding a custom error class, this at least allows ORMs to catch this specific error to override as they need.

I wasn't quite sure how to test this with the dummy_driver, but if a spec is needed, I can add one in.

### DIFF
--- a/src/db/error.cr
+++ b/src/db/error.cr
@@ -46,4 +46,9 @@ module DB
   # Raised when a scalar query returns no results.
   class NoResultsError < Error
   end
+
+  # Raised when the type returned for the column value
+  # does not match the type expected.
+  class ColumnTypeMismatchError < Error
+  end
 end
--- a/src/db/result_set.cr
+++ b/src/db/result_set.cr
@@ -80,7 +80,7 @@ module DB
       if value.is_a?(T)
         value
       else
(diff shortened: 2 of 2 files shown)


jwoertink:
Oh, I'd much rather it just tell you the specific column, but I figured between all of the different drivers, that would be a massive change.

Maybe we could do both? This change doesn't really break anything with any of the drivers, and would at least give a stepping stone in flexibility (or at least, that's my hope). But yeah, the best solution would be what you're suggesting @bcardiff

jwoertink:
Ah, I see what you're saying. I haven't looked too deep in to it, but we have `column_name(index : Int32)` already. If there's a way to know what the `current_index` is at the time of `read`, then it may be pretty simple.

jwoertink:
Yeah, in this case I defined the json column as `JSON::Any?`, but in Avram I defined it as `JSON::Any` (not nilable). Then when I pulled out the value and assign to the type, it tries to read as `column.read(JSON::Any)`, but the value was NULL.

I'm sure we can get it better, but this is already a HUGE step forward. I appreciate the help with this one. I say you merge in your changes, and we start getting more eyes on it.

PR #159, opened 2022-01-21
Returns block's value on #transaction method

### LINKED ISSUE #145: Transactions don't return the result of the yielded block

bcardiff:
Fixes #145.

Since 0.11 was never released yet, if wanted we can include this change also.

cc: @matthewmcgarvey

### DIFF
--- a/src/db/begin_transaction.cr
+++ b/src/db/begin_transaction.cr
@@ -11,7 +11,7 @@ module DB
     # The exception thrown is bubbled unless it is a `DB::Rollback`.
     # From the yielded object `Transaction#commit` or `Transaction#rollback`
     # can be called explicitly.
-    # Returns the value of the block when no exception was raised.
+    # Returns the value of the block.
     def transaction(& : Transaction -> T) : T? forall T
       tx = begin_transaction
       begin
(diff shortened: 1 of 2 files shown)


matthewmcgarvey (approved):
:shipit:

src/db/begin_transaction.cr
```diff
@@ -11,12 +11,14 @@ module DB
     # The exception thrown is bubbled unless it is a `DB::Rollback`.
     # From the yielded object `Transaction#commit` or `Transaction#rollback`
     # can be called explicitly.
-    def transaction
+    # Returns the value of the block when no exception was raised.
```
straight-shoota, line 14: Minor: The latter part is technically unnecessary. When the method raises, it generally doesn't have a return value.
```suggestion
    # Returns the value of the block.
```

straight-shoota (approved):

PR #168, opened 2022-04-18
Allow the use of enums

jgaskins:
I often want one or more properties of my DB-backed models to be constrained to a low-cardinality set of values, which can be known at compile time. Usually I end up storing them in strings, but then I can't lean on the Crystal type system to help me constrain those values throughout my application.

Since the `JSON` module supports Crystal enums, I looked into adding support for them here, as well. With this I can remove the translation layer between the HTTP layer and the DB layer if the only difference between them is an `enum`.

I was originally going to add a macro so you'd do something like `DB.register_type MyEnum`, but then I remembered that `Enum` is itself a Crystal type that represents all `enum`s to the exclusion of all else, which made it really simple to support without doing any extra work.

### DIFF
--- a/spec/serializable_spec.cr
+++ b/spec/serializable_spec.cr
@@ -90,10 +90,10 @@ struct ModelWithEnum
   getter c2 : MyOtherEnum
 
   enum MyEnum
-    Foo
-    Bar
-    Baz
-    Quux
+    Foo  = 0
+    Bar  = 1
+    Baz  = 2
+    Quux = 3
   end
 
   enum MyOtherEnum
@@ -196,7 +196,7 @@ describe "DB::Serializable" do
   end
 
   it "should initialize a model with an enum property" do
-    expect_model("1,Baz,LOL", ModelWithEnum, {
+    expect_model("1,2,LOL", ModelWithEnum, {
       c0: 1,
       c1: ModelWithEnum::MyEnum::Baz,
       c2: ModelWithEnum::MyOtherEnum::LOL,
--- a/spec/serializable_spec.cr
+++ b/spec/serializable_spec.cr
@@ -81,6 +81,29 @@ class ModelWithJSON
   property c1 : String
 end
 
+struct ModelWithEnum
+  include DB::Serializable
+
+  getter c0 : Int32
+  getter c1 : MyEnum
+  # Ensure multiple enum types work together
+  getter c2 : MyOtherEnum
+
+  enum MyEnum
+    Foo
+    Bar
+    Baz
+    Quux
+  end
+
+  enum MyOtherEnum
+    OMG
+    LOL
+    WTF
+    BBQ
+  end
+end
+
 macro from_dummy(query, type)
   with_dummy do |db|
     rs = db.query({{ query }})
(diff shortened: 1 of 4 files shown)


straight-shoota (approved):

jgaskins:
Found a bug in the error message while using this, so I just pushed up a couple commits to touch that up.

spec/serializable_spec.cr
```diff
@@ -172,6 +195,17 @@ describe "DB::Serializable" do
     expect_model("1,a", ModelWithJSON, {c0: 1, c1: "a"})
   end
 
+  it "should initialize a model with an enum property" do
+    expect_model("1,Baz,LOL", ModelWithEnum, {
```
beta-ziliani, line 199: It'd be nice to have a similar test but with the int values of the enum. @jgaskins do you want to take it, or do you prefer us to do it? Thanks for your PR!

jgaskins: @beta-ziliani Nice catch, I thought I included a test for that. I'll go ahead and add it.

jgaskins: Looks like this isn't as cut-and-dry as I thought it would be. The `DummyDriver` returns everything as strings, so the `case` statement always ends up in the `String` code path.

I tried modifying [this line](https://github.com/crystal-lang/crystal-db/blob/07c68d38e49e06ae4a56c4060da4bd2748df6701/spec/dummy_driver.cr#L190) to detect numbers (returns `n.to_i64? || n` instead of `n`), but it broke [this spec](https://github.com/crystal-lang/crystal-db/blob/07c68d38e49e06ae4a56c4060da4bd2748df6701/spec/database_spec.cr#L11). That's the only spec that breaks, but I can change that to a non-numeric value if that makes the most sense.

Any guidance on this one?

jgaskins: @beta-ziliani I added ae9a7a82505220399da88840bd9f0a0ab9e6925a to do this. Let me know if updating that other spec is too broad a change and I'll back it out.

beta-ziliani: Ugh, I guess we can put it as `pending` for the moment and leave it for the future to have a better stub for the ResultSet that includes column types. WDYT?

straight-shoota: @beta-ziliani With https://github.com/crystal-lang/crystal-db/commit/ae9a7a82505220399da88840bd9f0a0ab9e6925a this looks good to me, actually.

straight-shoota (approved):

spec/serializable_spec.cr
jgaskins: That makes sense. I'll revert https://github.com/crystal-lang/crystal-db/commit/ae9a7a82505220399da88840bd9f0a0ab9e6925a and add a pending spec for it.

straight-shoota: Wait. What's the problem with the current solution? I think it works. The number/string selection might not be very elegant, but IMO it's okay. We can't use strings that parse as integers in tests. But that's okay. There's no need to do that because we can chose the test values freely.

beta-ziliani: OK then. The mockup ResultSet is all about parsing strings after all, so I guess you're right. @jgaskins let's leave it as is.

beta-ziliani (approved):
Thanks @jgaskins ! 🚀

spec/serializable_spec.cr
jgaskins: Fantastic. I was concerned it might be too broad a change to modify the test harness to accommodate a tiny (though useful) feature, but I'm glad to know it's okay. Thanks friends!

straight-shoota: I suppose we'll have to see whether that breaks anything in other db drivers that use the same test harness.

PR #170, opened 2022-05-26
Add support for connection pool health checks

### LINKED ISSUE #169: Connection Pool does not check if connections have become disconnected
Sockets (TCP and Unix) are often disconnected by a firewall, docker, or system settings on various hosts (fly.io and render.com were the two ones I tested; fly.io uses TCP sockets and render uses Unix ones).  Because those disconnects were server-side, the client wasn't aware that the connections in the pool had become disconnected, and those connections will be treated as open sockets/connections. 

This is partially discussed in #109 and probably a few other places.

Here are steps to reproduce this error in a Postgre app (specifically killing individual connections). I was using Lucky+Avram, but happy to help reproduce this on another set of tools if necessary.

In a low-traffic production app on fly.io, I was seeing this error multiple times daily, and it would periodically force a server restart due to all connections being in a "disconnected" state with no ability to recover.

---

Replication:

Connect an app in dev, have a database setup with a maximum idle pool of 1, initial pool size of 1, max pool size 1.
Kill the connection directly in SQL (PG example): `SELECT pg_terminate_backend(pid) FROM pg_stat_activity WHERE pid <> pg_backend_pid() AND state = 'idle' LIMIT 1;` -- this kills a random idle connection, so if you have more current connections, you may need to run it a few times.

Wait a minute after executing that SQL command, then try to visit an endpoint in your app that would trigger a database request. Assuming you've killed all the connections, you should see a DB::ConnectionLost error be raised (true in PG, at the very least). This will especially be true if you have a relatively low connection timeout for the retry attempt.

robcole:
Resolves #169 

Provides an API to address disconnections to
long-running connections that occur outside code
by adding support for periodic health checks on idle
connections.

Periodic connection reapers cycle through a list of
idle connections, and call a #check method provided by the
connection. Check methods should either result in success on in
raising a PoolResourceLost(T) error such as DB::ConnectionLost.

Connections that fail a health check are replaced within the Connection Pool.

---

PR Notes:

Most pools in various languages handle this with rescuing the failed connection, ejecting it from the pool, and retrying -- this adds latency spikes and in my apps was actually triggering DB::ConnectionLost errors (probably due to retry_attempts + retry_delay not being able to get a successful retry in before timing out, but I'm not entirely certain).

Golang and Java both have ConnectionPools that introduce a TTL for connections and kill them. I initially experimented with that, but killing all idle connections, regardless of their current health, was fairly inefficient and created a large number of unnecessary allocations.

Instead of going with a TTL-based approach, I implemented an optional "health check" that adapters can add to their Connection class. https://github.com/the-business-factory/crystal-pg/pull/1 pairs with this PR 🍷 🧀 and implements a very simple health check.

In a demo low traffic app @ https://hirobot.app/ the health check itself has eliminated the disconnects, where earlier attempts at implementing TCP keepalive settings on PG, the server, and the database did not; it appears that some firewall or cluster-level disconnects are prevented by the periodic ACK.

### DIFF
--- a/src/db/connection.cr
+++ b/src/db/connection.cr
@@ -31,7 +31,11 @@ module DB
     # :nodoc:
     property auto_release : Bool = true
 
+    @created_at : Time
+    getter created_at
+
     def initialize(@context : ConnectionContext)
+      @created_at = Time.utc
       @prepared_statements = @context.prepared_statements?
     end
 
@@ -50,6 +54,9 @@ module DB
       raise DB::Error.new("There is an existing transaction in this connection") if @transaction
       @transaction = true
       create_transaction
+    end
+
+    protected def check
     end
 
     protected def create_transaction : Transaction
--- a/src/db/pool.cr
+++ b/src/db/pool.cr
@@ -18,6 +18,10 @@ module DB
     @retry_attempts : Int32
     # seconds to wait before a retry attempt
     @retry_delay : Float64
+    # time to wait before beginning connection pool health checks
+    @reaping_delay : Float64
+    # amount of time in between pool health checks
+    @reaping_frequency : Float64
 
     # Pool state
 
@@ -37,14 +41,73 @@ module DB
     # global pool mutex
     @mutex : Mutex
 
-    def initialize(@initial_pool_size = 1, @max_pool_size = 0, @max_idle_pool_size = 1, @checkout_timeout = 5.0,
-                   @retry_attempts = 1, @retry_delay = 0.2, &@factory : -> T)
+    def initialize(
+      @initial_pool_size = 1,
+      @max_pool_size = 0,
+      @max_idle_pool_size = 1,
+      @checkout_timeout = 5.0,
+      @retry_attempts = 1,
+      @retry_delay = 0.2,
+      @reaping_delay = 60.0,
+      @reaping_frequency = 60.0,
+      &@factory : -> T
+    )
       @availability_channel = Channel(Nil).new
       @waiting_resource = 0
       @inflight = 0
       @mutex = Mutex.new
 
       @initial_pool_size.times { build_resource }
+    end
+
+    def reap_connection!(connection : T, delay : Float64)
+      return unless connection.responds_to?(:check)
+
+      sleep delay
+
+      sync do
+        connection_idle = @idle.includes? connection
+        return unless connection_idle # let the next reaping attempt get it
+
+        delete connection
+      end
+
+      connection.check
+
+      sync do
+        @total << connection
+        @idle << connection
+      end
+    rescue exc : DB::ConnectionLost
+      connection.close
+
+      sync do
+        add_resource! if can_increase_pool?
+      end
+    end
+
+    def add_resource!
+      @inflight += 1
+      r = unsync { build_resource }
+    ensure # prevent inflight from incrementing if build_resource fails
+      @inflight -= 1
+      r
+    end
+
+    def start_reaper!
+      spawn do
+        sleep @reaping_delay
+
+        loop do
+          sync do
+            @idle.each_with_index do |connection, i|
+              spawn reap_connection!(connection, i * 0.1)
+            end
+          end
+
+          sleep @reaping_frequency
+        end
+      end
     end
 
     # close all resources in the pool
@@ -77,10 +140,7 @@ module DB
         until resource
           resource = if @idle.empty?
                        if can_increase_pool?
-                         @inflight += 1
-                         r = unsync { build_resource }
-                         @inflight -= 1
-                         r
+                         add_resource!
                        else
                          unsync { wait_for_available }
                          # The wait for available can unlock
@@ -206,6 +266,11 @@ module DB
     # :nodoc:
     def is_available?(resource : T)
       @idle.includes?(resource)
+    end
+
+    # :nodoc:
+    def is_in_pool?(resource : T)
+      @total.includes?(resource)
     end
 
     # :nodoc:
--- a/spec/pool_spec.cr
+++ b/spec/pool_spec.cr
@@ -220,4 +220,63 @@ describe DB::Pool do
 
     all.size.should eq 4
   end
+
+  it "should close idle resources after a health check fails" do
+    DummyDriver::DummyConnection.clear_connections
+    DB.open "dummy://localhost:1027?initial_pool_size=1&max_pool_size=1&reaping_delay=0.0&reaping_frequency=60.0" do |db|
+      cnn = db.checkout.as(DummyDriver::DummyConnection)
+      cnn.closed?.should be_false
+      cnn.release
+      db.pool.is_available?(cnn).should be_true
+      db.pool.is_in_pool?(cnn).should be_true
+      cnn.disconnect!
+      sleep(0.05)
+      db.pool.is_available?(cnn).should be_false
+      db.pool.is_in_pool?(cnn).should be_false
+    end
+  end
+
+  it "should introduce a small delay between health checks" do
+    DummyDriver::DummyConnection.clear_connections
+    DB.open "dummy://localhost:1027?initial_pool_size=2&max_pool_size=3&max_idle_pool_size=3&reaping_delay=0.0&reaping_frequency=60.0" do |db|
+      cnn = db.checkout.as(DummyDriver::DummyConnection)
+      cnn2 = db.checkout.as(DummyDriver::DummyConnection)
+      cnn.release
+      cnn2.release
+      cnn.disconnect!
+      cnn2.disconnect!
+      db.pool.is_available?(cnn).should be_true
+      db.pool.is_in_pool?(cnn).should be_true
+      db.pool.is_available?(cnn2).should be_true
+      db.pool.is_in_pool?(cnn2).should be_true
+      sleep(0.05)
+      db.pool.is_available?(cnn).should be_false
+      db.pool.is_in_pool?(cnn).should be_false
+      db.pool.is_available?(cnn2).should be_true
+      db.pool.is_in_pool?(cnn2).should be_true
+      sleep(0.05)
+      db.pool.is_available?(cnn2).should be_false
+      db.pool.is_in_pool?(cnn2).should be_false
+    end
+  end
+
+  it "should not close open resources" do
+    DummyDriver::DummyConnection.clear_connections
+    DB.open "dummy://localhost:1027?initial_pool_size=1&max_pool_size=2&reaping_delay=0.0&reaping_frequency=60.0" do |db|
+      cnn = db.checkout.as(DummyDriver::DummyConnection)
+      cnn2 = db.checkout.as(DummyDriver::DummyConnection)
+      cnn.release
+      cnn.disconnect!
+      cnn2.disconnect!
+      db.pool.is_available?(cnn).should be_true
+      db.pool.is_in_pool?(cnn).should be_true
+      db.pool.is_available?(cnn2).should be_false
+      db.pool.is_in_pool?(cnn2).should be_true
+      sleep(0.05)
+      db.pool.is_available?(cnn).should be_false
+      db.pool.is_in_pool?(cnn).should be_false
+      db.pool.is_available?(cnn2).should be_false
+      db.pool.is_in_pool?(cnn2).should be_true
+    end
+  end
 end
--- a/src/db/database.cr
+++ b/src/db/database.cr
@@ -59,6 +59,7 @@ module DB
         @setup_connection.call conn
         conn
       }
+      @pool.start_reaper!
     end
 
     # Run the specified block every time a new connection is established, yielding the new connection
--- a/src/db/driver.cr
+++ b/src/db/driver.cr
@@ -36,6 +36,8 @@ module DB
         checkout_timeout:   params.fetch("checkout_timeout", 5.0).to_f,
         retry_attempts:     params.fetch("retry_attempts", 1).to_i,
         retry_delay:        params.fetch("retry_delay", 1.0).to_f,
+        reaping_frequency:  params.fetch("reaping_frequency", 60.0).to_f,
+        reaping_delay:      params.fetch("reaping_delay", 60.0).to_f,
       }
     end
   end


src/db/connection.cr
```diff
@@ -31,7 +31,11 @@ module DB
     # :nodoc:
     property auto_release : Bool = true
 
+    @created_at : Time
+    getter created_at
+
     def initialize(@context : ConnectionContext)
+      @created_at = Time.utc
```
robcole, line 38: I added this + the getter initially for some very helpful debugging information as I was trying to address system-level disconnects and figure out when they were beginning to happen; I'm not using it currently and could remove it, but it's turned out to be useful enough that I opted to leave it for review as a possible useful addition.

src/db/pool.cr
```diff
@@ -47,6 +60,56 @@ module DB
...
+    end
+
+    def start_reaper!
+      spawn do
+        sleep @reaping_delay
+
+        loop do
+          sync do
+            @idle.each_with_index do |connection, i|
+              spawn reap_connection!(connection, i * 0.1)
```
robcole, line 104: The idea with the 100ms stacking delay here is to prevent clearing out all of the `@idle` connections at the same time.

src/db/pool.cr
```diff
@@ -47,6 +60,56 @@ module DB
...
+        delete connection
+      end
+
+      connection.check
+
+      sync do
+        @total << connection
+        @idle << connection
+      end
+    rescue exc : DB::ConnectionLost
```
robcole, line 81: Should probably be PoolResourceLost(T)

src/db/connection.cr
```diff
@@ -52,6 +56,9 @@ module DB
       create_transaction
     end
 
+    protected def check
```
straight-shoota, line 59: Could you add some documentation to this method, please? Just a short description what implementing drivers should do here.

src/db/pool.cr
```diff
@@ -47,6 +60,56 @@ module DB
...
+    ensure # prevent inflight from incrementing if build_resource fails
+      @inflight -= 1
+      r
+    end
+
+    def start_reaper!
+      spawn do
+        sleep @reaping_delay
+
+        loop do
```
straight-shoota, line 101: This fiber should be stopped when the database gets closed.

src/db/pool.cr
straight-shoota: Could we check the individual connections consecutively instead of concurrently?
Spawning a new fiber for each connection seems unnecessary.

src/db/pool.cr
```diff
@@ -47,6 +60,56 @@ module DB
       @initial_pool_size.times { build_resource }
     end
 
+    def reap_connection!(connection : T, delay : Float64)
+      return unless connection.responds_to?(:check)
```
straight-shoota, line 64: `Connection` implements `#check`, so this guard seems to be unnecessary? However, we could consider to remove the stub implementation and let it be implemented only when it actually does something.

src/db/pool.cr
straight-shoota: Actually, every fiber acquires a lock in `reap_connection!`, so they would execute consecutively anyways.

src/db/pool.cr
```diff
@@ -47,6 +60,56 @@ module DB
...
+        add_resource! if can_increase_pool?
+      end
+    end
+
+    def add_resource!
+      @inflight += 1
+      r = unsync { build_resource }
+    ensure # prevent inflight from incrementing if build_resource fails
+      @inflight -= 1
+      r
```
straight-shoota, line 94: Caching the return value is unnecessary. The method returns the value of the main body, not the `ensure` block.
```suggestion
      unsync { build_resource }
    ensure # prevent inflight from incrementing if build_resource fails
      @inflight -= 1
```

src/db/pool.cr
grepsedawk: I think this spawn is just a form of creating a timer here

src/db/connection.cr
robcole: 👍

src/db/pool.cr
robcole: Yeah, this was intended to be a timer.

My goal here was to create a timer that injected 100ms of time in between attempts to remove an `@idle` connection -- the goal here was to attempt to make sure that the `@idle` pool always retains a spare connection (aside from when using `max_idle_pool_size=1`).

straight-shoota: Sure. But your code essentially executes equivalent to this plain flow. 

```cr
@idle.each_with_index do |connection, i|
  reap_connection!(connection, i * 0.1)
end
```

The fibers don't actually run concurrently because of the mutex. That also wrecks the delay. All fibers sleep as soon as they are spawned, then they block waiting for the mutex (except the first one). When the previous one releases the mutex, the next executes directly because it's already waiting. So the delay is without effect.

robcole: That makes sense, thanks.

Given that, what do you think about something like this? The main thing I'm not certain of is whether the health check should live inside of the mutex (my guess is these should usually resolve in ~100ms, and aside from PG, not many adapters may choose to implement this) -- if so, the implementation can be really clean, as the health check itself would act as the delay:

```crystal
    def start_reaper!
      spawn do
        sleep @reaping_delay

        loop do
          @idle.each { |connection| reap_connection! connection }
          sleep @reaping_frequency
        end
      end
    end
    
    def reap_connection!(connection : T)
      sync do
        if @idle.includes? connection
          delete connection
          connection.check
          @total << connection
          @idle << connection
        end
      end
    rescue exc : PoolResourceLost(T)
      connection.close
      sync { add_resource! if can_increase_pool? }
    end

```

src/db/pool.cr
robcole: It feels like `sleep @reaping_frequency` would throw that loop off, so I may still be missing something here -- this was inspired by https://lbarasti.com/post/select_statement/ but I definitely don't fully grok it yet.

src/db/pool.cr
straight-shoota: I don't think the health check should be synced. It includes an network request, hence there is no guarantee how long it might take to complete.

src/db/pool.cr
straight-shoota: You can use `when timeout(@reaping_frequency)` instead of `else`. This means: What happens first, the timeout completes or the channel receives, the respective branch executes.

src/db/pool.cr
robcole: Makes sense -- updated accordingly.

src/db/connection.cr
robcole: Done!

src/db/pool.cr
```diff
@@ -34,24 +38,77 @@ module DB
...
+    def start_reaper!
+      spawn(name: "connection_reaper") do
+        sleep @reaping_delay
+
+        loop do
+          select
+          when @close_channel.receive?
+            break
+          when timeout @reaping_frequency.seconds
+            @idle.each { |connection| reap_connection! connection }
```
robcole, line 100: revisit to see if a delay method is worthwhile adding back w/the current implementation

src/db/pool.cr
robcole: Done, thanks!

robcole:
@straight-shoota Thanks for the review -- recent commits have some updated that should address the feedback so far. I could add a test around ensuring that the `connection_reaper` fiber is exited, but all of the ways I can think of to test that require a bit of extra state, so let me know if you think it's warranted.

Any other feedback is welcome -- I've got the latest commit up and running on https://hirobot.app, and everything looks great with it in production.

src/db/pool.cr
```diff
@@ -34,24 +38,79 @@ module DB
...
+
+      connection.check
+
+      sync do
+        @total << connection
+        @idle << connection
+      end
+    rescue exc : PoolResourceLost(T)
+      connection.close
+      sync { add_resource! if can_increase_pool? }
```
straight-shoota, line 83: Does this rescue block need to span the entire method? I suppose it could only cover the `connection.check` call, right?

robcole: The connection needs to be added back to @total and @idle on success, so I can't quite see an alternative way to write the rescue to only span the `connection.check` call -- but definitely open to ideas as the only area where PoolResourceLost would be raised is in the connection.close call.

src/db/pool.cr
```diff
@@ -34,24 +38,79 @@ module DB
...
+    def start_reaper!
+      spawn(name: "connection_reaper") do
+        sleep @reaping_delay
+
+        loop do
+          select
+          when @close_channel.receive?
+            break
+          when timeout @reaping_frequency.seconds
+            @idle.each_with_index { |conn, i| reap_connection! conn, i * 0.1 }
```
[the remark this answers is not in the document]
robcole: Makes sense, adjusted. 👍

robcole:
Let me know if there's anything I should do to help with the (seemingly non-deterministic?) test failures in CI -- I'm guessing that the sleep + timing checks may be a bit too restrictive (?) and failing due to that, but I'm not 100% certain.

Tests had no issue on OSX locally w/1.4.1 installed, happy to do what I can locally to get this rock-solid if there is anything.

src/db/pool.cr
robcole: @straight-shoota LMK if you have a suggestion for this one. 👍

robcole:
Just pushed up a commit to try to help resolve the nondeterministic test failures, but since they're only occurring in CI it's hard to determine if I'm making things better or not.

src/db/pool.cr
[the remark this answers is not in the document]
straight-shoota: I actually meant moving the call to `sleep` out of `reap_connection!`, next to the call.
The suggestion was actually flawed, this should better show it:

```suggestion
            @idle.each_with_index do |conn, i|
              sleep i * 0.1
              reap_connection! conn
            end
```

src/db/pool.cr
```diff
@@ -34,21 +38,49 @@ module DB
...
+    def start_reaper!
+      spawn(name: "connection_reaper") do
+        loop do
+          select
+          when @close_channel.receive?
+            break
+          when timeout @reaping_frequency.seconds
+            sync do
+              @idle.each_with_index { |conn, i| reap_connection! conn, i }
+            end
```
straight-shoota, line 75: I don't think syncing the entire reaper sequence is going to work. `sync` acquires a lock at the entry and only releases it on exit. So while the block is running, all other sync operations are blocked (they'll be unblocked at `unsync`, but that's only a minor relief). Most of the time in the reaper loop is spent sleeping and there is definitely no need to sync that.
Syncs should cover each `reap_connection!` individually and without the sleep.

robcole: 👍 that makes sense -- I reverted the commit, but then added a bit to a test that will help illustrate the problem I was looking to fix with the current implementation. 

Let me know if you have any suggestions, but the gist of the issue is that the sleep isn't actually doing anything in terms of enforcing the delay to prevent the `@idle` pool from being exhausted while waiting on health checks.

robcole: https://github.com/crystal-lang/crystal-db/pull/170/commits/56e0a35c18f32f8d1e360504976baad2b191f151 shows the issue I was trying to get around, but I can see that my approach isn't correct for that.

I added some puts debugging to track which "attempt" the reaper is on when things are being checked, which helps me see where some of the timing-dependent tests were. 

LMK if you have any immediate thoughts, but I may need to rethink my approach a bit here -- my only real goal is to prevent having all the `@idle` connections be removed at the same time. 

I'm going to give a shot at moving things into a `@health_check` pool (or possibly just using a buffered channel?), but I'm open to any other ideas to help simplify the code and make the testing story a bit more reliable.

robcole: Definitely a bit of a 🤦 moment.

Code is much simpler now, and I *think* the tests should be much less brittle, or at the very least, much easier to pad to help slower CI boxes compensate now.

src/db/pool.cr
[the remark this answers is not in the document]
robcole: 👍 -- see other comments for why it turns out sleeping isn't necessary at all, so I removed all of this.

robcole:
Submitted some updates that should make the tests a lot less brittle. If they fail again, I'm curious if it may have something to do with the precision of the sleep statements I'm using in the tests, so I can try to pad them a bit next.

robcole:
Increased the test sleep statements a bit more to see if this gets tests running in CI more reliably. 

Going to see if I can get these working on my repo as well so I don't have to keep pinging y'all with "hmm, is this better?" for the CI failures.

robcole:
Last commit seems to have things very stable based on the repeated runs I've tried @ https://github.com/the-business-factory/crystal-db/actions/workflows/ci.yml

straight-shoota (approved):

robcole:
@straight-shoota I'm a little uncomfortable with the flakiness of these tests because they're so timing-dependent -- they mostly seem to be passing, but even after tweaking everything I've had a couple failed runs @ https://github.com/the-business-factory/crystal-db/actions/runs/2484059819 -- my guess is that this just has to do with the precision of when a slept fiber ends up running (have seen a few issues discussing this as well) but I don't have any clear ideas for how to resolve it fully.

LMK if you think there's any way they can be improved, or if any of them should be removed (the `reaper_delay` one is particularly problematic).

robacarp:
@straight-shoota what do you think this pull request needs in order to be a candidate for the next release?

jgaskins (commented):
> In a demo low traffic app @ https://hirobot.app/ the health check itself has eliminated the disconnects, where earlier attempts at implementing TCP keepalive settings on PG, the server, and the database did not; it appears that some firewall or cluster-level disconnects are prevented by the periodic ACK.

This looks like a shortcoming of the `Socket` implementation. Keepalives are implemented in libc/kernel code but the `closed?` state is based on [an instance variable in `Socket`](https://github.com/crystal-lang/crystal/blob/994c70b10e9939d73b561135b0822dea4f9979bb/src/socket.cr#L439-L441) which AFAICT is not set when it fails keepalives.

If we can fix that it would probably handle most of the use cases for explicit health checks.

robcole:
Thanks for the feedback so far, @jgaskins - I’m going to take some time to look into Socket as I was as under the impression that this was a OS-level event that needed to be handled at the Pool layer to avoid connections that have become stranded.

wout:
What's the progress on this? I'd love to see this merged (and I'm not alone).

xaviablaza:
Would love to see this merged in!

robcole:
> 

I haven't had enough time lately to incorporate the feedback from @straight-shoota and @jgaskins, most of which I 100% agree with. I do think that this belongs as a core feature and would appreciate the help in making a better testing story for it. 

It's been running on my https://hirobot.app for quite a while now and has drastically increased stability, so let me know if there's anything I can do on (for now, limited time) to help get this merged as well, #crystal-lang team.

beta-ziliani:
We're currently focusing on the 1.8 release of the compiler.  I'm adding it to our internal tracking and come back to you after the release.

jgaskins:
FWIW, I still hold the position I mentioned in [the related issue](https://github.com/crystal-lang/crystal-db/issues/169#issuecomment-1264511904) — the connections themselves should be self-healing rather than being polled by the connection pool. If that's not possible, the connection can [close itself and the pool will remove it](https://github.com/crystal-lang/crystal-db/blob/87dc8aafaf1fe5acc2c2203ea7610d0949bda207/src/db/pool.cr#L144-L145).

Self-healing connections avoid the following problems:

- the race condition I mentioned above regarding connection count
- the extra resources required to check every connection in the pool
- timing deviations with `reaping_frequency` I mentioned above
- exceptions that can still occur between the failure and the check interval — if a connection can't self-heal, you can still get a single exception per failed connection, but it won't go on for up to a full minute by default

Implementing self-healing connections can be simple in a lot of cases, especially with [ACID](https://en.wikipedia.org/wiki/ACID)-compliant databases — single queries and `transaction` blocks can both simply be retried N times with reconnection attempts in between. A couple real-world examples of this are [Neo4j](https://github.com/jgaskins/neo4j.cr/blob/9e91a1d87b81b0c4de2f6e0146529f5cf65cbd12/src/neo4j/bolt/connection.cr#L618-L629) and [Redis](https://github.com/jgaskins/redis/blob/337fdd51b23be1fc25670af103e0d746dee39ad7/src/connection.cr#L351-L368) (not ACID, but atomicity is the most important feature for this). Both shards simply call `initialize` if there's an `IO::Error`, which reconnects to the server, before retrying the query or transaction.

I've never encountered any connection issues with either of these databases in production since implementing self-healing connections in them (both use `DB::Pool` for pooling), and users of my Redis shard have told me they haven't had any connection issues since migrating to it from `stefanwille/redis`, which does not implement self-healing connections.

jgaskins:
Hmm, it looks like `DB::Pool` [already attempts retries and throws away any broken or refused connections](https://github.com/crystal-lang/crystal-db/blob/87dc8aafaf1fe5acc2c2203ea7610d0949bda207/src/db/pool.cr#L163-L195) out of the box (any time you invoke a query/exec method on `DB::Database`, it calls this method via [`DB::PoolStatement#statement_with_retry`](https://github.com/crystal-lang/crystal-db/blob/0415deebbb5174ffb98385f9037834f0c6a6dc86/src/db/pool_statement.cr#L41-L45)). If this isn't working effectively, and it sounds like it isn't, that sounds like a bug. Could the driver be missing a `raise ConnectionLost.new` somewhere? Or is this reproducible on another DB (in which case the bug could be in this shard)?

robcole:
> Hmm, it looks like `DB::Pool` [already attempts retries and throws away any broken or refused connections](https://github.com/crystal-lang/crystal-db/blob/87dc8aafaf1fe5acc2c2203ea7610d0949bda207/src/db/pool.cr#L163-L195) out of the box (any time you invoke a query/exec method on `DB::Database`, it calls this method via [`DB::PoolStatement#statement_with_retry`](https://github.com/crystal-lang/crystal-db/blob/0415deebbb5174ffb98385f9037834f0c6a6dc86/src/db/pool_statement.cr#L41-L45)). If this isn't working effectively, and it sounds like it isn't, that sounds like a bug. Could the driver be missing a `raise ConnectionLost.new` somewhere? Or is this reproducible on another DB (in which case the bug could be in this shard)?

Wish I could dedicate more time to giving you the response depth you need, but I'm crushed time-wise right now (planning a move to a new state and pretty slammed with work).

The tl;dr of what I observed is that when a connection is interrupted in an idle state -- which in a low-traffic app that has TCP-based timeouts (common in most containerized setups, like fly.io and render.com, etc) is pretty often -- is that eventually, the entire HTTP pool becomes saturated with dead TCP connections (closed by the other side).

Once the connection pool is fully saturated with dead connections, the retry counts and delays start to back up, and connections just fail, or responses that need to happen quickly (this is the case on webhook acknowledgement and registration w/Slack) take too long and cause downstream issues.

As far as I'm aware, there's no way to "heal" a TCP connection that is killed outside the app's knowledge; you have to remove it from the pool and add a new connection. That's the main angle I was trying to pursue but never found a solution for, and if there is something that would allow that, I could look into.

#169 gives a pretty kludgy (but the best I could do, given the constraints of how this error shows up) way of replicating this, by killing an underlying Postgre process on a DB with a limited pool.

@jgaskins A final note -- this happens super reliably on not only my app, but also on quite a few people I've chatted with in the Lucky and Crystal discord channels (a few have let me know they're running my patch and it's helped resolve their issues). 

I do think that this validates that it's a real problem and that active pool management is an approach worth considering, but like I said, I'm very open to other ways to solve this, and may not be able to meet the technical requirements you're looking for in a solution.

jgaskins:
To be clear, I'm not implying that this isn't a problem to be solved. My questions at the bottom of my previous comment were to find out where the bug is in the existing implementation — `crystal-pg` or `crystal-db`. If this is reproducible with `crystal-mysql`, then it's likely that the bug exists in this shard. However, I feel like it might be in `crystal-pg` because detecting a dirty disconnection over TCP is a hard problem.

> As far as I'm aware, there's no way to "heal" a TCP connection that is killed outside the app's knowledge; you have to remove it from the pool and add a new connection.

This is true, but `DB::Pool` doesn't deal with the `TCPSocket` directly. It deals with `DB::Connection`. 🙂 The driver's connection implementation can throw away the socket and spin up a new one, which would heal the driver's connection without throwing it away. For example, in my `redis` shard, `DB::Pool` doesn't throw away the `Redis::Connection`, but the `Redis::Connection` throws away the `TCPSocket`. This lets `DB::Pool` avoid having to know about the connection's health — other than whether or not it's closed.

> I do think that this validates that it's a real problem

Since I noticed the code in my previous comment, I don't think the connection necessarily needs to self-heal because `DB::Database` queries already go through a self-healing code path built into this shard (which is, admittedly, not very well documented). It looks like it just needs to raise `DB::ConnectionLost` and `DB::Pool` will try to fix it. Since that's not working for you, I agree that that's definitely a real problem, but it's a bug somewhere rather than a need for a new solution for the same problem.

bcardiff:
Self healing connections sounds nice, but I fear that that pushes the complexity to the drivers. If we want to go that route an alternative is to split the pool resource and the protocol implementation. Also there are related objects like prepared statements that depends on connection instances which need to be dropped / evicted from the corresponding caches if the connection object wants to be reused.

The db pool currently retries on a connection lost, this mostly handles scenarios where db goes down as shown in https://crystal-lang.org/reference/1.8/database/connection_pool.html . But also this retry is only active when you exec/query directly on the database, not when dealing with single connection.

This last bit also means that when initiating a transaction or checking out explicitly a connection there is no retry involved. Might this be the case where the web apps that are suffering ends up?

Another benefit of going with the TTL approach is to release memory from prepared statements by the database. AFAIK in heroku this was an issue sometimes and partially why we have `prepared_statements=false` as an alternative. Also Go sql implemented TTL after I inspired crystal-db pool design in it.

The mention of idle connections being the problem of a saturated pool make me wonder if what we need is a `max_checkout_time`. In case the resource is never returned to the pool because it's leaked we can close it and make room again in the pool. This can be done without an external fiber and is a bit more deterministic.

beta-ziliani:
If I'm following the discussion right, there are currently three alternatives:
 1. Do nothing; let drivers' internals handle it, maybe [fixing the Socket implementation](https://github.com/crystal-lang/crystal-db/pull/170#pullrequestreview-1127571327).
 2. Implement this PR, fixing [some of the issues raised](https://github.com/crystal-lang/crystal-db/pull/170#issuecomment-1494834990) and implementing a way to opt-out for drivers that do not need it (without wasting resources).
 3. Implement another approach, like TTL.
 
A combination of the first and any of the other two is likely desirable. Am I missing something? Thoughts?

PR #178, opened 2023-03-27
Fix mt issues

bcardiff:
Thanks to https://github.com/crystal-lang/crystal-sqlite3/issues/87 some issues with the pool and statement cache for mt were discovered.

This PR address the known issues.

Some more might come in a later PR if we narrow it.

### DIFF
--- a/src/db/pool_prepared_statement.cr
+++ b/src/db/pool_prepared_statement.cr
@@ -35,7 +35,11 @@ module DB
     # the connection is registered in `@connections`
     private def build_statement : Statement
       clean_connections
-      conn, existing = @db.checkout_some(@connections)
+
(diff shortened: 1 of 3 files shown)


src/db/pool_prepared_statement.cr
```diff
@@ -39,16 +42,22 @@ module DB
         conn.release
         raise ex
       end
-      @connections << WeakRef.new(conn) unless existing
+      unless existing
```
beta-ziliani, line 45: Can line 38 above be problematic? we pass unguarded the set of connections to `checkout_some`, but it can change mid-way, can't it?

beta-ziliani (approved):

straight-shoota (approved):

PR #179, opened 2023-04-10
Add io_provider to DB.open / DB.connect

### LINKED ISSUE #162: Using DB with an already existing socket or connection
I want to make it possible to use the pg driver with an [`SSH2::Channel`](https://github.com/spider-gazelle/ssh2.cr/blob/1589903e0284e533169af854777225a0eaa29772/src/channel.cr). It's possible, I have it working by using the pg driver directly without crystal-db, and with monkey patching `PQ::Connection` to take the ssh channel as a socket. So that's all cool, but it'd be nice to also use crystal-db. 

In general it may also be nice to let people establish their own tcp socket ahead of time and pass it in. Maybe? 

I think for these uncommon cases it'd be okay to make people create a postgres connection manually, and pass it into crystal-db to wrap. This may already be possible and I just haven't figured out how yet. If not what do you think the best way to do this would be?

bcardiff:
Fixes #162 .

Drivers implementation should call `ConnectionContext#io_provider : DB::IOProvider?` method to optionally use the the specified provider.

Since not all drivers are IO based is hard to make this parameter mandatory.

I would like to discuss what should be a nice default. Maybe providing TCPSocket or UnixSocket builders directly based on the hostname as crystal-pg does. But for dice each driver would need some default_port at least since that is driver specific. Maybe the default io_provider should also be driver specific and DB provide a base implementation.

### DIFF
--- a/spec/dummy_driver.cr
+++ b/spec/dummy_driver.cr
@@ -9,6 +9,7 @@ class DummyDriver < DB::Driver
   class DummyConnection < DB::Connection
     def initialize(context)
       super(context)
+      context.io_provider.try &.build_io
       @connected = true
       @@connections ||= [] of DummyConnection
       @@connections.not_nil! << self
--- a/spec/io_provider_spec.cr
+++ b/spec/io_provider_spec.cr
@@ -1,0 +1,54 @@
+require "./spec_helper"
+require "log/spec"
+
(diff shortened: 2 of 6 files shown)


bcardiff:
I have another hopefully better alternative to share.

PR #180, opened 2023-05-29
Use https link for API docs

jgaskins:
Linking to [the https site](https://crystal-lang.github.io/crystal-db/api/latest/) mostly to avoid the "Not Secure" warning in my browser, but also because it's a good idea

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ Note: Multiple drivers can be included in the same application.
 
 ## Documentation
 
-* [Latest API](http://crystal-lang.github.io/crystal-db/api/latest/)
+* [Latest API](https://crystal-lang.github.io/crystal-db/api/latest/)
 * [Crystal book](https://crystal-lang.org/docs/database/)
 
 ## Usage


bcardiff (approved):

bcardiff:
I've enabled "Enforce HTTPS" that generates an http -> https redirect on GHP.

Still jekyll-redirect-from seems to generate redirects with http. If the browser still complains we might get around it setting up the `site.url` property according to https://github.com/jekyll/jekyll-redirect-from#prefix so the redirect will be to an https url.

> If site.url is set, its value, together with site.baseurl, is used as a prefix for the redirect url automatically. This is useful for scenarios where a site isn't available from the domain root, so the redirects point to the correct path. If site.url is not set, only site.baseurl is used, if set.
>
> Note: If you are hosting your Jekyll site on [GitHub Pages](https://pages.github.com/), and site.url is not set, the prefix is set to the pages domain name i.e. http://example.github.io/project or a custom CNAME.

jgaskins:
Thanks @bcardiff! Between these two changes, it looks like the link in the readme goes to an `https` site now. 💯

PR #186, opened 2023-07-20
Fix max_idle_pool_size race condition

### LINKED ISSUE #182: `max_idle_pool_size` usage causes "disappearing" fibers
Under load, when actively using the db pool some fibers "disappear". I assume they are blocked somewhere. In the repo below we provide two reproductions - one using a simple web interface and load tested using ab (Apache Benchmark) and one using a simple script that just spawns a bunch of fibers and tries to use the pool.

This only happens when fiddling with the `max_idle_pool_size` param. In this example we want a fixed pool, hence `initial_pool_size=5&max_pool_size=5&max_idle_pool_size=5`. See the repo for details on how to reproduce this.

This has already been [discussed in the forum](https://forum.crystal-lang.org/t/concurrency-issues-with-an-api-server/5482) a while ago.

[Github repo for reproduction](https://github.com/emil-rb/crystal-db-pool-issue)

bcardiff:
Fixes #182

I changed the DummyDriver to simulate IO by introducing a couple of Fiber.yield along the way. This allowed me to extract the load test from #182 without external dependencies and be able to reproduce the race condition. I'm adding these in spec/manual to avoid loosing it.

After fixing the race condition some cleanup lead to dropping the `@waiting_resource` which is good since it implied knowing a bit too much about was happening all the time.

I think the fixed race condition can be stressed by 2 pool clients with a max_pool_size=1 and max_idle_pool_size=1 by the following trace.

- Client A: performs a checkout and eventually yields
- Client B: performs a checkout, the pool is at max capacity and needs to wait, so increments the `@waiting_resource++`, and blocks on the `@availability_channel.receive`
- A: performs a release. `@idle << resource` and wants to `@availability_channel.send`
- Scheduler: the send cause a yield but client b could also experience a timeout
- B: the universe throws a coin and choose to perform a timeout (there is no guarantee what would win)
- B: `@waiting_resource--`, a timeout is raised
- B: The `retry` method which was invoking the `checkout` makes a retry. `@idle` is not empty and can be poped.

B continues with the resource while A is blocked on the `@availability_channel.send`

By using non-blocking sends and dropping the `@waiting_resource` we are embracing better what could happen.

### DIFF
--- a/spec/manual/load_test.cr
+++ b/spec/manual/load_test.cr
@@ -1,0 +1,54 @@
+# This file is to be executed as:
+#
+# % crystal ./spec/manual/load_test.cr
+#
+# It generated a number of producers and consumers. If the process hangs
+# it means that the connection pool is not working properly. Likely a race condition.
+#
+
+require "../dummy_driver"
+require "../../src/db"
+require "json"
+
+CONNECTION = "dummy://host?initial_pool_size=5&max_pool_size=5&max_idle_pool_size=5"
+
+alias TChannel = Channel(Int32)
+alias TDone = Channel(Bool)
+
+COUNT = 200
+
+def start_consumer(channel : TChannel, done : TDone)
+  spawn do
+    indeces = Set(Int32).new
+    loop do
+      indeces << channel.receive
+      puts "Received size=#{indeces.size}"
+      break if indeces.size == COUNT
+    end
+    done.send true
+  end
+end
+
+def start_producers(channel : TChannel)
+  db = DB.open CONNECTION do |db|
+    sql = "1,title,description,2023 " * 100_000
+
+    COUNT.times do |index|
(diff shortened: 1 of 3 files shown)


spec/manual/load_test.cr
```diff
@@ -0,0 +1,54 @@
+# This file is to be executed as:
+#
+# % crystal ./spec/manual/load_test.cr
+#
+# It generated a number of producers and consumers. If the process hangs
```
beta-ziliani, line 5:
```suggestion
# It generates a number of producers and consumers. If the process hangs
```

beta-ziliani (commented):
LGTM! And less blocking too!

beta-ziliani (approved):

beta-ziliani (approved):

PR #196, opened 2023-11-09
Deprecate `DB.mapping`

straight-shoota:
[`DB::Serializable`](https://crystal-lang.github.io/crystal-db/api/0.12.0/DB/Serializable.html) has been around for a long time and is a more versatile tool.
The equivalents `JSON.mapping` and `YAML.mapping` in stdlib are long gone and replaced by `*::Serializable`.

cf. https://github.com/crystal-lang/crystal-db/pull/195#issuecomment-1802215887


bcardiff (approved):
I wouldn't mind removing it directly and creating a community shard for allowing a migration path.

PR #207, opened 2024-06-21
Allow `Enumerable` for query args instead of `Array`

lwakefield:
This is a little bit of a "see if it sticks", in the sense that I don't have a full grasp of what might break as a result of this proposed change (even though specs pass!)

The impetus for this change, is getting named bindings to work. On top of these changes, I can now do the following:

```crystal
class SQLite3::Statement
    private def bind_arg(index, tuple : Tuple)
        k, v = tuple
        idx = LibSQLite3.bind_parameter_index(self, ":"+k.to_s)
        bind_arg(idx, v)
    end
end

db.query_all("select 'hello '||:msg", args: {:msg => "world!"}, as: {String})
```

So - WDYT?

### DIFF
--- a/src/db/pool_statement.cr
+++ b/src/db/pool_statement.cr
@@ -15,7 +15,7 @@ module DB
     end
 
     # See `QueryMethods#exec`
-    def exec(*args_, args : Array? = nil) : ExecResult
+    def exec(*args_, args : Enumerable? = nil) : ExecResult
       statement_with_retry &.exec(*args_, args: args)
     end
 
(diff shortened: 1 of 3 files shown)


bcardiff (approved):
Regarding lifting Array to Enumerable it seems time has come https://github.com/crystal-lang/crystal-db/pull/110#issuecomment-531817680 :-) 

On the named argument api though I am not so sure if treating Tuples like named argument is the accurate thing. Sure is convenient, but what if you want to send an actual tuple as value? An alternative would be to have a separate record `NamedArgument` which can be implemented by each driver. The api become less attractive though

```crystal
db.query_all("select 'hello '||:msg", args: [NamedArgument(:msg, "world!")], as: {String})
```

Maybe for convenience we can have a named arg `named_args` that is implemented on top of the above construct (or not).

```crystal
db.query_all("select 'hello '||:msg", named_args: {msg: "world!"}, as: {String})
```

straight-shoota (approved):

PR #209, opened 2024-10-11
`crystal tool format` with Crystal 1.15.0-dev

straight-shoota:
Crystal nightly has a couple new formatter rules enabled: https://github.com/crystal-lang/crystal/pull/14718
The changes are backwards-compatible, so Crystal 1.14.0 won't complain about the new format.

### DIFF
--- a/spec/dummy_driver.cr
+++ b/spec/dummy_driver.cr
@@ -319,13 +319,13 @@ class Witness
   end
 end
 
-def with_witness(count = 1)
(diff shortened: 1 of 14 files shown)


bcardiff (approved):
Late to the party but formatter forcing all those & does not look nice to me.

straight-shoota:
Well I suppose it's not set in stone that we *definitely* roll this out. Let's see how it fares.

PR #211, opened 2024-11-28
Replaced deprecated splat operator

BigBoyBarney:
Came across this deprecation warning when running `crystal tool dependencies` in another projects

```cr
In lib/db/src/db.cr:92:26

 92 | alias Any = Union({{*TYPES}})
                           ^
Warning: Deprecated use of splat operator. Use `#splat` instead
```


straight-shoota (approved):

bcardiff:
I was worried if merging this required bumping the minimal crystal version, but it doesn't need to. 😁

PR #216, opened 2025-05-23
Add ability to expire resources within the pool

### LINKED ISSUE #47: Add an option to check for liveness on checkout
Feature request:  AFAICT there is no concept of "is this connection still alive" in the connection pool system.

Adding something like that might be nice. The easiest way (on my head) I can think of to implement this would be "at checkout time" it performs some trivial check like "select 1" or what not, before returning a connection, or something similar.

Thanks!

syeopite:
Closes #47

This PR implements the ability to expire assets within the pool based on what I think was discussed in #47

This is configured through the two new pool parameters `max_lifetime_per_resource` and `max_idle_time_per_resource` which sets the number of seconds before they get expired and removed from the pool. This is done on each `#checkout` and `#release`  but also with an optional background fiber which runs on either a configured timer, or whichever configured expiration is shorter.

This PR mostly also respects `initial_pool_size` in that it'll try to keep at least that many resources within the pool--but it doesn't actually know which resources are stale until it checks on a `#checkout/#release` or with the background fiber so they're not necessarily always fresh.

But if any of the expiration check causes the amount of resources within pool to drop below `initial_pool_size` it'll immediately replenish with fresh resources till its back up to `initial_pool_size`.

Also when a `#checkout` internally finds an expired resource, the new `PoolResourceExpired` exception is raised. So `#retry` will get more important.

The specs all seems to pass and modifying the manual specs to use the new expiration features did not show any glaring problems either. 

Please let me know if I've missed something with this implementation! I wrote this over the course of a single night so apologies in advance if the code might be a bit messy 😅

### DIFF
--- a/src/db/pool.cr
+++ b/src/db/pool.cr
@@ -175,6 +175,7 @@ module DB
       @total.each &.close
       @total.clear
       @idle.clear
+      @resource_lifecycle.clear
       @sweep_job_close_channel.send(nil) if @sweep_job_running
     end
 
--- a/src/db/pool.cr
+++ b/src/db/pool.cr
@@ -109,6 +109,10 @@ module DB
 
     # whether the job is enabled or disabled
     @sweep_job_enabled : Bool
+    # has a sweep job running
+    @sweep_job_running : Bool
+    # timer between each run
+    @sweep_timer : Time::Span?
     # cancels the sweep job as needed
     @sweep_job_close_channel : Channel(Nil)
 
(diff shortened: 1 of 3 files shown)


src/db/pool.cr
```diff
@@ -282,5 +445,66 @@ module DB
...
+
+    # Ensure there are at least a minimum of @initial_pool_size non-expired resources
+    #
+    # Should be called after each expiration batch
+    private def ensure_minimum_fresh_resources
+      replenish = (@initial_pool_size - (@total.size + @inflight)).clamp(0, @initial_pool_size)
+
+      return if replenish <= 0
+
+      replenish.times do |index|
```
syeopite, line 495: The resource replenishment could actually block a checkout/release operation for a really long time depending on how many resources it needs to be built. In most cases it should just be a single one but it can quickly balloon when many resources are discarded manually from outside of the pool. I'm not entirely sure what's the best way to resolve this.

I feel like spawning a new fiber each time is excessive, so maybe it could be dispatched to the already existing background job?

There's also the fact that any non-blocking operation will remove the guarantee of there always being at least `initial_pool_size` fresh resources available...

PR #226, opened 2026-02-18
Check whether pool resource was closed in `before_checkout` hook, allowing for health checks and max-TTL connections

### LINKED ISSUE #225: Allow expiring old resources
I remember seeing discussions around setting a max TTL on resources. I can't remember where any of those discussions are, but I can't find them in this repo. Please feel free to close if I've simply overlooked them.

One thought I had while playing around with this idea last night was letting the resource type itself handle expiration using `before_checkout`. Basically, `checkout` runs `before_checkout` and then checks `resource.closed?`. If it's closed, it drops the connection and repeats the loop.

Another idea was to add a `before_release` hook. I have a branch where this is a 3-line change (plus the specs):

```diff
diff --git a/src/db/pool.cr b/src/db/pool.cr
index 017b35c..8441bc1 100644
--- a/src/db/pool.cr
+++ b/src/db/pool.cr
@@ -162,6 +162,9 @@ module DB
       idle_pushed = false

       sync do
+        if resource.responds_to?(:before_release)
+          resource.before_release
+        end
         if resource.responds_to?(:closed?) && resource.closed?
           @total.delete(resource)
         elsif can_increase_idle_pool
```

Unfortunately, while the upside is a trivial implementation, the downside of relying on this for expiration is that it places the check after you're done using it, so you have to use the "expired" connection at least once.

jgaskins:
This PR allows pool resources to do things like run health checks before returning from `checkout` to make sure the connection is healthy:

```crystal
class PG::Connection
  getter last_health_check = Time.instant

  def before_checkout
    if last_health_check.elapsed > 1.minute
      exec "SELECT 1" 
      @last_health_check = Time.instant
    end
  rescue ex : IO::Error
    close
  end
end
```

Shopify does something similar to this. At the beginning of a request, they ping the MySQL server (via [the `trilogy` adapter](https://github.com/trilogy-libraries/trilogy)) to ensure the connection is alive and, if not, they explicitly close it, which causes ActiveRecord's connection pool to open a new one, possibly to a different replica entirely.

This PR can also be used to allow connections to have a max TTL. The driver's `DB::Connection` subclass can have a `before_checkout` that checks against a TTL. For example:

```crystal
class PG::Connection
  getter opened_at = Time.instant
  property ttl = 1.hour

  def before_checkout
    close if opened_at.elapsed > ttl
  end
end
```

Fixes #225

### DIFF
--- a/spec/pool_spec.cr
+++ b/spec/pool_spec.cr
@@ -57,6 +57,19 @@ class Closable
   end
 end
 
+class Expirable
+  include DB::Disposable
+
+  property? expired = false
+
+  def before_checkout
+    close if expired?
+  end
+
+  private def do_close
+  end
+end
+
 private def create_pool(**options, &factory : -> T) forall T
   DB::Pool.new(DB::Pool::Options.new(**options), &factory)
 end
@@ -192,6 +205,19 @@ describe DB::Pool do
     all[0].closed?.should be_false
     all[1].closed?.should be_true
     all[2].closed?.should be_false
+  end
+
+  it "should not return closed resources from checkout" do
+    pool = create_pool(max_pool_size: 1, max_idle_pool_size: 1) { Expirable.new }
+
+    resource1 = pool.checkout
+    pool.release resource1
+
+    # Expired while idle
+    resource1.expired = true
+
+    resource2 = pool.checkout
+    resource2.should_not eq resource1
   end
 
   it "should not return closed resources to the pool" do
--- a/src/db/pool.cr
+++ b/src/db/pool.cr
@@ -135,16 +135,24 @@ module DB
                      else
                        pick_available
                      end
-        end
 
-        @idle.delete resource
+          if resource
+            if resource.responds_to?(:before_checkout)
+              resource.before_checkout
+            end
+
+            @idle.delete resource
+
+            if resource.responds_to?(:closed?) && resource.closed?
+              @total.delete resource
+              resource = nil
+            end
+          end
+        end
 
         resource
       end
 
-      if res.responds_to?(:before_checkout)
-        res.before_checkout
-      end
       res
     end
 


drum445:
I agree, we need something like this. Currently I would argue the DB libraries in crystal cannot be used in prod, as if the site is idle for N hours, all the connections in the pool are dead and can't recover.

bcardiff:
Sorry @jgaskins i never left feedback for this.

I think this looks in a good direction. I'm ok with a responds_to being used.

I think we should be explicit on what is the protocol though. Should the method close the resource? What happens if an exception is thrown? Specs don't cover this.

I think that either raising or resource being closed are good contracts. 

If that's sound good we should have a couple of more specs and docs.

jgaskins:
In re-reading this, I realized I set the PR title incorrectly due to some confusion on my part. The `before_checkout` already exists [on `master`](https://github.com/crystal-lang/crystal-db/blob/29bf17d0351df55770d57e2540d5a99af1bd0de0/src/db/pool.cr#L145-L147), but if it results in closing the connection, `DB::Pool#checkout` doesn't detect it and will try to send a query on the closed connection.

The best-case scenario when this happens is that this falls back on the `retry` logic, which adds `retry_delay` seconds of latency per connection. But if the connection is severed uncleanly, we run into situations where we simply don't receive data we're expecting from the server and wait for the socket to time out at the TCP layer, which could be anywhere from several seconds to forever.

I'll update the PR title to reflect what the changes actually do.

> I think we should be explicit on what is the protocol though. Should the method close the resource?

Agreed. I'm not confident enough to prescribe specifically what `before_checkout` *should* do here, unfortunately. I only know that I want to use it for: improving reliability and latency by closing connections that fail a health check.

So I think the method should be *allowed to* close the connection, but what else it does feels like it might be implementation-dependent.

> What happens if an exception is thrown? Specs don't cover this.

This is a good point. I would've assumed the original specs around `before_checkout` would have covered this, but I'm not seeing anything in there about it. Do you have suggestions?

ysbaddaden:
About raise: whatever the hook did with the connection failed hard, so I guess the resource ought to be removed.

Checking `.closed?` after the hook returns is a little weird, but given that we handle connections, I think it's fine to assume connections to have a closed state. Maybe checkin could also check whether the connection has been closed and cleanup instead of returning?

PR #231, opened 2026-08-20
Added Mimer SQL to the list

majorproblem:
I have written a driver for the relational database [Mimer SQL](https://mimer.com) that uses crystal-db, and I would like to add that to the list of drivers.


straight-shoota (approved):

bcardiff:
Thanks! So encouraging to have someone from mimer maintaining the crystal driver.

